PR #2, opened 2019-04-19
add trailing comma to prettier config

kdv24:
- This fixes the removal of trailing comments on save, but I haven't yet found a way to prevent vscode making multi-line arrays into single line. 
- https://github.com/prettier/prettier-vscode/issues/352

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -2,5 +2,6 @@
   "jsxBracketSameLine": true,
   "printWidth": 80,
   "semi": true,
-  "singleQuote": true
+  "singleQuote": true,
+  "trailingComma": "all"
 }


kdv24:
Thanks for the quick review, @thmsdnnr.  I'll see what @dougmakingstuff thinks. I'm not sure who else is from the Amplify Team for sure to ask.

bobbymart1n:
This works for me. I do have format on save set to true on my editor. I'm still a little new to VS Code and have been using Atom for the past year. As far as I'm aware, VS Code looks for a .prettierrc file if you have prettier installed which I do currently. Let me know if I'm mistaken about that though as I think I've run into VS Code having its own format on save functionality and overriding the .prettierrc file in the past.

kdv24:
@bobbymart1n I'm fairly new to the VS Code settings as well.  I have a .prettierrc file and format on save set for me as well, but as I understand it the multi-line issue with the array isn't something prettier will do... or at least I'm not having any luck with it.  For example, in `src/store.js` when I save it, my editor reformats
```
const middleware = [
  thunk,
  routerMiddleware(history),
];
```
to ` const middleweare =[thunk, routerMiddleware(history)];` all on one line.

If you change yours to multiline and save it, does it revert it back to the multi-line?  I'm just trying to determine if it's autoformatting for you.

Thanks!

PR #33, opened 2019-10-23
disable auth on master

augustinevt:
NOTE: MAYBE MERGE TO A NEW `development` BRANCH INSTEAD OF `master`? OR KEEP VERSIONS OF THESE FILES SO WE DON'T RUN INTO TROUBLE REACTIVATING AUTH LATER?

Summary:
- Replace/remove conditional auth logic in `Authorization.jsx` so that the wrapped component is always returned.

- Replace/remove conditional auth logic in `PrivateRoute.jsx` so that the wrapped component is always returned.

BEFORE:
testGif

AFTER:
testGif


trezp:
Woooo YAY!!!!! 👏 👏 👏

astuteape:
This is awesome! Thank you @augustinevt and @trezp. I was able to merge this locally with `develop` and navigate between pages like you show in the gif. @augustinevt I think you're right that it's a good idea to merge this with develop rather than master. Maybe we can get into a git-flow practice that way. Would you mind opening a new PR to `develop`? I'll merge it in and update in Slack that `develop` is the branch we should all work from on our local forks, if that sounds good to you two.

PR #45, opened 2019-12-05
Master

htharker42:
moved old app into a folder labeled stashed
created new react project 
added react-router-dom 
customized landing page


ThomasShetler:
When is next meetup?

PR #59, opened 2020-01-30
added assets from old projct, moved logos to correct location in asse…

htharker42:
Added Assets folder and old assets from old project. Moved existing logos to appropriate locations in assets folder and updated imports.


htharker42:
merging as there is no real code to review

PR #116, opened 2020-03-31
Add property page

### LINKED ISSUE #97: Create Add Property page
(See page 15 of mockups: https://drive.google.com/file/d/1YqboQogczYm1HkyRqEtVSzeQ61T9hWU2/view)

- [x] Use POST `/properties` endpoint with parameters:
- name
- address
- city
- zipcode
- state
(?) Do we need to add a `units` parameter to the backend to match the mockup?
- [x] Create the Add Property page behind a PrivateRoute
- [x] Add Formik form with proper Yup validation

zfregin:
Add property page component within a protected route. Contains Formik form with Yup validation to post form data to /properties api. Things to review include:
- [x] Styling
- [ ] Validation
- [ ] Successful POST

Still need to review the need to adjust the backend endpoint to accept units as a parameter. Will it accept an assigned property manager in addition to this?


dwellingly/src/App.js
```diff
@@ -87,6 +88,7 @@ export class App extends React.Component {
               <PrivateRoute exact path='/' component={Dashboard} />
               <Route exact path='/login' component={LoginForm} />
               <PrivateRoute exact path='/dashboard' component={Dashboard} />
+              <Route exact path='/add-property' component={AddProperty} />
```
Chris-Boe: Since we should have persistent auth merged in soon, we should make this a "PrivateRoute" since you need to be authenticated to use it

dwellingly/src/views/addProperty.js
```diff
@@ -0,0 +1,162 @@
...
+                                            <Field
+                                                className="column form-field"
+                                                type="text"
+                                                name="units"
+                                                onChange={handleChange}
+                                                value={values.zipcode}
+                                                placeholder="10"
+                                            />
+                                        </div>
+                                        <div className=" add-property__assign-manager-container">
```
Chris-Boe: Like we talked about, let's comment this out and make a new issue for it. Be sure to mention the `react-select` package in the issue.
It'll be waiting on #120 to get done, then we can start using that api endpoint.

dwellingly/src/views/addProperty.js
```diff
@@ -0,0 +1,162 @@
+import React, { Component } from 'react';
+import { Form, Field, Formik } from 'formik';
+import * as Yup from 'yup';
+import { UserContext } from '../App';
+import * as axios from 'axios';
+
+const validationSchema = Yup.object().shape({
+    name: Yup.string()
+        .min(5, "*Name must have at least 2 characters")
+        .max(50, "*Names can't be longer than 100 characters")
```
Chris-Boe: Getting mixed signals here 😄 Sounds like the min should be 2 and the max should be 100

dwellingly/src/views/addProperty.js
```diff
@@ -0,0 +1,162 @@
...
+        .required("*Name is required"),
+    address: Yup.string()
+        .required("*Address is required"),
+    city: Yup.string()
+        .required("*City is required"),
+    zipcode: Yup.string()
+        .required("*Zipcode is required"),
+    state: Yup.string()
+        .required("*State is required")
```
Chris-Boe, line 23: We'll want to keep in mind in the future a general "Address schema" that we can check forms against to make sure we have common types for things like state (OR or Oregon?), Zip (5 digits? or 5 digits + the postal extension?), etc.
Not sure what that will look like yet so we don't need to worry about it here

Chris-Boe (changes requested):
Nice and sleek!

dwellingly/src/views/addProperty.js
```diff
@@ -0,0 +1,162 @@
...
+                                            {errors.zipcode ? (<div className="error-message">{errors.zipcode}</div>) : null}
+                                        </div>
+                                        <div className="form-row columns">
+                                            <label className="column is-one-fifth" htmlFor="zipcode">Units</label>
+                                            <Field
+                                                className="column form-field"
+                                                type="text"
+                                                name="units"
+                                                onChange={handleChange}
+                                                value={values.zipcode}
```
Chris-Boe, line 144: Looks like this line has a copy/paste issue.
I noticed when I type in the 'zip' form element, it adds my input to the "Unit" box, and I can't type into the Unit box directly.

image

dwellingly/src/views/addProperty.js
```diff
@@ -0,0 +1,162 @@
...
+    city: Yup.string()
+        .required("*City is required"),
+    zipcode: Yup.string()
+        .required("*Zipcode is required"),
+    state: Yup.string()
+        .required("*State is required")
+});
+
+const formHandler = (data, context) => {
+    axios.post('http://localhost:5000/properties', data, { headers: {"Authorization" : `Bearer ${context.user.accessJwt}`} })
```
Chris-Boe, line 27: This _looks_ right but I'm getting a 422 with an error saying it's missing the Bearer token. I think "Authorization" should _not_ be in quotes by what I can find online

dwellingly/src/views/addProperty.js
```diff
@@ -0,0 +1,162 @@
...
+        .then(function(response){
+            console.log(response);
```
Chris-Boe: We'll want to make a Toast or something for both the OnSuccess handling and the Onfailure handling, but for now let's switch them to alert(response) so we can see what goes on

Chris-Boe (changes requested):

dwellingly/src/App.js
```diff
@@ -90,6 +91,7 @@ export class App extends React.Component {
               <PrivateRoute exact path='/' component={Dashboard} />
               <Route exact path='/login' component={LoginForm} />
               <PrivateRoute exact path='/dashboard' component={Dashboard} />
+              <PrivateRoute exact path='/add-property' component={AddProperty} />
```
Chris-Boe: We have the `/add/property` PrivateRoute in the list below from the left nav, so we won't need this one

dwellingly/src/views/addProperty.js
```diff
@@ -0,0 +1,172 @@
...
+        .max(100, "*Names can't be longer than 100 characters")
+        .required("*Name is required"),
+    address: Yup.string()
+        .max(250, "*Address can't be longer than 250 characters")
+        .required("*Address is required"),
+    city: Yup.string()
+        .max(50, "*City can't be longer than 50 characters")
+        .required("*City is required"),
+    zipcode: Yup.string()
+        .max(50, "*City can't be longer than 20 characters")
```
Chris-Boe: Couple of copy/paste errors in here and the state validation below

Chris-Boe (changes requested):

Chris-Boe (approved):

PR #164, opened 2020-04-23
Move logout button to top right.

### LINKED ISSUE #152: move logout button
move logout button to upper right hand corner

wrunion:


stashed/src/components/user-controls/UserControls.scss
```diff
@@ -1,8 +1,8 @@
 .user-controls {
   color: #ffffff;
   position: absolute;
-  bottom: 0px;
-  left: 0px;
+  top: 0px;
```
Chris-Boe, line 4: So it looks like you are trying to change files in the stashed/ folder 😄 This is actually the old version of the code that is no longer being used, we just have kept it around in case we need to pick out and reuse any helpful components or code. The only code you'll need to be working on is what's in the dwellinglyapp/src/ folder

Chris-Boe (changes requested):
Also be sure to name your branch something that portrays what work is being done on it, like `move-logout-button` or `issue-152`.
Thanks!

stashed/src/components/user-controls/UserControls.scss
wrunion: Hi Chris! So I took a look at the current code, and if I'm seeing correctly, it looks like moving the button is gonna take actual component refactoring, instead of just changing a CSS rule. The LogOutButton component is currently in Nav, and it looks like I'd have to move it to Header - which I'd be happy to do, but since I'm still new to this code base, I'm hesitant to make a change that large. 

Maybe this is a better issue for someone who's more familiar with the structure of the app? I just don't want to introduce bugs by moving a component to a place people aren't expecting. Or if you think it's safe to make the changes, I'm happy to do that too. 

Thanks!

wrunion: Alright I moved the logout button into the Header component, which should make styling easier. I tried to write new styles but something else must be overriding them.

It's where it needs to be now though, component-wise, and I made a note of where the new CSS rules need to go. I'll read up on Burma and SASS, but I wanted to get this request moving along, and see if anyone else knows which rules are affecting which components, and how to override them for this.

Alright!

Chris-Boe: ~I don't see your changes, were you going to wait on committing them until the CSS was figured out? Feel free to reach out for assistance!~ Oh nevermind! I see your other PR. Closing this one.
And don't feel hesitant to move big things and break stuff 😉  that's what feature branches are for and it's a great way to understanding the codebase better! 😄

PR #172, opened 2020-05-07
List users pending access for accept/decline. #163

### LINKED ISSUE #163: Integrate Dashboard/Request for Access with the Backend
The dashboard page has a list for new users who don't have their access level assigned yet. These users will be triaged by an admin and their access will then be established. (Otherwise, newly registered users have their access set to "pending")

- Find users with pending access by: POST /users/role and provide a body with "userrole": "pending"
- If the admin selects "DECLINE" and then "YES", delete the user by id with: DELETE /user/<id>
  - Repeat the step to find users with pending access and use that data to refresh the list
- If the admin selects "ADD", go to the "Request for Access" page

chrisandrewca:
Part of resolve #163

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4343,7 +4343,8 @@
         },
         "kind-of": {
           "version": "6.0.2",
-          "resolved": ""
+          "resolved": "https://registry.npmjs.org/kind-of/-/kind-of-6.0.2.tgz",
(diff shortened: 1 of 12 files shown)


Chris-Boe (approved):
Looks great with the backend's #34 PR!

Issue #174, opened 2020-05-07
Integrate tenants needing staff assignments (dashboard) with backend


DaveTrost:
The dashboard has a space to list tenants with no staff assigned to them. This view allows JOIN admins to assign staff members from a drop-down. After clicking "Save Assignments" the tenants association to those staff members is updated via sending updates to the backend.

- [x] get a list of the tenants who have no staff assignments. GET from /tenants and those tenants without an assigned staff member will have `staff: None` in their JSON data
- [x] If the above tenants list contains unassigned staff, get a list of the JOIN staff members for assignment. POST to /users/role and include the following JSON object in the body: "userrole": "JOIN-Staff".
- [x] when the "Save Assignments" button is clicked, for each tenant whose staff assignment has been updated, send a PUT request to /tenants/<tenantID> and with a JSON body as follows: staffIDs: [<staffID from the staff members list>]
- [x] after the "Save Assignments" button is clicked, remove those tenants from the list who now have been given staff assignments

PR #176, opened 2020-05-13
Integrate tenants to dashboard

DaveTrost:
addressing issue #174. 

**NOTE: this code will run only with the backend branch "tenants-routes" (or with the dev branch after the tenants-routes PR is merged)**

- [x] get a list of the tenants who have no staff assignments. GET from /tenants and those tenants without an assigned staff member will have `staff: None` in their JSON data
- [x] get a list of the JOIN staff members for assignment. POST to /users/role and include the following JSON object in the body: "userrole": "admin".
- [x] when the "Save Assignments" button is clicked, for each tenant whose staff assignment has been updated, send a PUT request to /tenants/<tenantID> and with a JSON body as follows: staffIDs: [<staffID from the staff members list>]
- [x] after the "Save Assignments" button is clicked, remove those tenants from the list who now have been given staff assignments

### DIFF
--- a/src/views/dashboard.js
+++ b/src/views/dashboard.js
@@ -80,11 +80,12 @@ export const Dashboard = (props) => {
                 { 'staffIDs': [staff] }, 
                 makeAuthHeaders(session)
             ));
+
         axios.all(tenantUpdateReqs)
             .then(axios.spread((...responses) => {
-                const updatedTenants = unstaffedTenants.filter(({ id }) => {
-                    const isTenantUpdated = responses.find(({ data }) => data.id === id);
-                    // using filter to keep all but this updated tenant, so negate the result
+                const updatedTenants = unstaffedTenants.filter(tenant => {
+                    const isTenantUpdated = responses.find(({ data }) => data.id === tenant.id);
+                    // using filter to keep non-updated tenants, so negate the result
                     return !isTenantUpdated;
                 });
                 setUnstaffedTenants(updatedTenants);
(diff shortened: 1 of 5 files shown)


src/views/dashboard.js
```diff
@@ -1,15 +1,45 @@
...
+              const unstaffed = data.tenants.filter(tenant => !tenant.staff);
+              setUnstaffedTenants(unstaffed);
+          })
+          .catch((error) => {
+              alert(error);
+              console.log(error);
+          })
+      const data = { "userrole": "admin" };
+      axios
+          .post(`${process.env.REACT_APP_API_URL}/users/role`, data, makeAuthHeaders(session))
```
Chris-Boe: Since I don't see this data being used anywhere except in the dropdown to assign tenants, we could put this behind a conditional inside the "then" of the /tenants call above, so that we only call it if we have tenants that need staff assignment

Chris-Boe (changes requested):
Looks great! Just the one change and some merge conflicts

src/views/dashboard.js
DaveTrost: Good catch. I also had a non-empty dependency list in my useEffect and so the API calls were running twice on first load. I fixed this using a pattern called useMountEffect that I found on stack overflow

DaveTrost:
All issues noted in reviews should be handled now. Please take a look

Chris-Boe (approved):

PR #182, opened 2020-05-20
Footer update

mgrienauer:
Updated footer to appear at bottom of page and aligned text right with padding of 24px. The inner text was originally wrapped in a container but I removed it because it didn't look like it was matching the style of the Figma file. I can undo the padding and put it back in the container though.

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,20 +1,20 @@
-@import './scss/main.scss';
-@import './scss/pages/all';
+@import "./scss/main.scss";
+@import "./scss/pages/all";
 
 .App {
-  height: 100vh;
+	height: 100vh;
 }
-  
+
 .container {
-  flex-grow: 1;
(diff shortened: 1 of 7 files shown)


Chris-Boe (changes requested):
Just gotta get rid of all the extra files/changes 😄

mgrienauer:
Ok, I used git rebase and removed that init commit that seemed to be changing all those files. I think that I did it correctly, but mind taking a look? First time using rebase.

Chris-Boe:
I'm on kind of a wide-screen but I'm seeing a bit of a gap between the footer and the left nav when i have the browser at full screen
image

mgrienauer:
Ah, ok I gotta take another look at it tomorrow.

Chris-Boe (approved):

PR #177, opened 2020-05-13
Allow for frontend to be deployed on Heroku

KentShikama:
See https://dwellinglypdx.herokuapp.com for an example.

This uses `--buildpack mars/create-react-app`.

Note that all endpoints are now prefixed under `/api/` to match the backend (see https://github.com/codeforpdx/dwellinglybackend/pull/37).

PR #185, opened 2020-05-27
Tenant list

### LINKED ISSUE #146: Create Tenant List page
Page 8 of mockups https://drive.google.com/file/d/1YqboQogczYm1HkyRqEtVSzeQ61T9hWU2/view

zfregin:
Tenant List page, basic table layout to display tenant data

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -6,6 +6,7 @@ import { Home } from './views/home';
 import { NavMenu } from './components/NavigationMenu/navigationMenu.js';
 import { Dashboard } from './views/dashboard';
 import { Properties } from './views/properties';
+import { Tenants } from './views/tenants';
 import { Terms } from './views/terms';
 import { PrivateRoute, auth } from './Auth';
 import Header from './components/Header/index';
@@ -141,7 +142,7 @@ export class App extends React.Component {
               <PrivateRoute exact path='/add/tenant' component={Dashboard} />
               <PrivateRoute exact path='/add/property' component={AddProperty}/>
               <PrivateRoute exact path='/add/manager' component={Dashboard} />
-              <PrivateRoute exact path='/manage/tenants' component={Dashboard} />
+              <PrivateRoute exact path='/manage/tenants' component={Tenants} />
               <PrivateRoute exact path='/manage/properties' component={Properties} />
               <PrivateRoute exact path='/manage/managers' component={Dashboard} />
               <PrivateRoute exact path='/tickets' component={Dashboard} />
--- a/src/views/properties.js
+++ b/src/views/properties.js
@@ -67,7 +67,7 @@ export class Properties extends Component {
     }
 
     getProperties = (context) => {
-        axios.get(`${process.env.REACT_APP_API_URL}/properties`, { headers: {"Authorization" : `Bearer ${context.user.accessJwt}`} })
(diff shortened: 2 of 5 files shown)


src/views/properties.js
```diff
@@ -89,7 +89,7 @@ export class Properties extends Component {
                                 <Link className="button is-rounded" to="/add/property">+ ADD NEW</Link>
                             </div>
                             <div className="search-section">
-                                <input></input>
+                              <input class="input search is-rounded" placeholder="Search properties by name, address, or property manager"></input>
```
Chris-Boe: the "class" prop here will have to be "className" instead for React to pick it up

src/views/tenants.js
```diff
@@ -0,0 +1,138 @@
...
+              {session => {
+                  this.context = session;
+                  return (
+                      <div className="properties__container">
+                          <div className="section-header">
+                              <h2 className="page-title">Tenants</h2>
+                              <Link className="button is-rounded" to="/add/tenant">+ ADD NEW</Link>
+                          </div>
+                          <div className="search-section">
+                              <input class="input search is-rounded" placeholder="Search Tenants by name, property, or JOIN staff"></input>
```
Chris-Boe: "class" will have to be "className" here as well

src/views/tenants.js
```diff
@@ -0,0 +1,138 @@
+import React, { Component } from 'react';
+import BootstrapTable from 'react-bootstrap-table-next';
+import paginationFactory from 'react-bootstrap-table2-paginator';
```
Chris-Boe: I'm getting an error trying to run this because 'react-bootstrap-table2-paginator' is missing from package.json, so we'll have to add it there first

Chris-Boe (changes requested):

src/views/properties.js
zfregin: Another sloppy copy/paste from Bulma, but mostly using this as a dummy input until we build the real search component.

src/views/tenants.js
```diff
@@ -0,0 +1,138 @@
...
+
+        // this.getTenants = this.getTenants.bind(this)
+    }    
+
+    componentDidMount() {
+        // this.getTenants(this.context);
+    }
+
+    // getTenants = (context) => {
+    //     axios.get(`${process.env.REACT_APP_API_URL}/tenants`, { headers: {"Authorization" : `Bearer ${context.user.accessJwt}`} })
```
KentShikama: Note the axios requests will need to look like `axios.get("/api/tenants", ...` (see https://github.com/codeforpdx/dwellingly-app/pull/177)

src/App.js
```diff
@@ -147,6 +148,7 @@ export class App extends React.Component {
               <PrivateRoute exact path='/add/tenant' component={Dashboard} />
               <PrivateRoute exact path='/add/property' component={AddProperty}/>
               <PrivateRoute exact path='/add/manager' component={Dashboard} />
+              <PrivateRoute exact path='/manage/tenants' component={Tenants} />
```
Chris-Boe, line 151: Hmm there's already a /manage/tenants on line 154 here. Technically this one will get hit first, but we should remove the duplicate when we hook up this page to the backend

Chris-Boe (approved):

PR #186, opened 2020-05-27
Ticket list

### LINKED ISSUE #149: Create Ticket List page
Page 20 of the mockups https://drive.google.com/file/d/1YqboQogczYm1HkyRqEtVSzeQ61T9hWU2/view

zfregin:
Basic table layout to display tickets


src/views/tickets.js
```diff
@@ -0,0 +1,163 @@
...
+            <UserContext.Consumer>
+                {session => {
+                    this.context = session;
+                    return (
+                        <div className="tickets__container">
+                            <div className="section-header">
+                                <h2 className="page-title">Tickets</h2>
+                            </div>
+                            <div className="search-section">
+                              <input class="input is-rounded" placeholder="Search by tenant, manager, property, or JOIN staff"></input>
```
Chris-Boe, line 117: don't forget to use "className" throughout this file instead of "class" so react can use it

Chris-Boe (changes requested):
Looks good so far! Had a few issues with class vs className to fix. I didn't test with any real data, since Hugh is working on the backend endpoints for Tickets currently, but if anything breaks we can come back to it when that gets in!
I noticed the table doesn't fill the page horizontally like in the mockup, but the style fixes might end up fixing that: 
image

src/views/tickets.js
zfregin: Yeah, I just copy pasted from Bulma and forgot to make the change. Sounds like we will be building out some of these primitives ourselves though

src/views/tickets.js
```diff
@@ -0,0 +1,163 @@
...
+
+        this.getTickets = this.getTickets.bind(this)
+    }    
+
+    componentDidMount() {
+        this.getTickets(this.context);
+    }
+
+    getTickets = (context) => {
+        axios.get(`${process.env.REACT_APP_API_URL}/properties`, { headers: {"Authorization" : `Bearer ${context.user.accessJwt}`} })
```
KentShikama: Note the axios requests will need to look like `axios.get("/api/properties", ...)` (see https://github.com/codeforpdx/dwellingly-app/pull/177)

Chris-Boe:
Let's also take out the changes to `dwellingly/package-lock.json` since that file no longer exists and we don't want to add it back in

Chris-Boe (approved):
👍

PR #191, opened 2020-06-06
Add Procfile for dokku builds

KentShikama:
This is required for dwellingly to seamlessly work with Dokku. See https://github.com/mars/create-react-app-buildpack/issues/54#issuecomment-310111156 .

See https://dwellingly.distributed.directory/ for a running example.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,1 @@
+web: bin/boot


Chris-Boe:
We aren't using docker or dokku, and instead we're deployed on Heroku. What do we need this for?

KentShikama:
It would be nice to be able to run this application on Dokku without having to maintain a fork in which this file is added. And if in the future you decide to switch hosts like how RecordSponge did, it will be easier. This project has already moved from using Firebase, so it wouldn't be surprising if it had to change hosting providers in the future. Otherwise, if there is little value in having a mirror site like https://dwellingly.distributed.directory/ (for demoing/testing purposes) or if there is little chance that this application will be used in a different city/context, then this is not so useful.

KentShikama:
Note this doesn't interfere with deployment to Heroku if that is a concern.

Chris-Boe (approved):

Issue #48, opened 2020-01-16
Build Front-End Forms/CRUD for user creation


htharker42:
connect to backend api
create, updated, destroy

skrabacher:
Model variable names here: https://github.com/codeforpdx/dwellinglyapi/blob/master/models/user.py

PR #215, opened 2020-06-18
Add registration page v3

spensbot:
This solves issue #48. I need to incorporate the standard notification popup/dialog to notify of successful/failed registrations before redirecting to login.

I added a css class to "typography.scss" for the form errors. Let me know if that's not appropriate.

Thanks for making my 2nd pull request possible!


Chris-Boe (approved):
Nice!

Issue #206, opened 2020-06-15
FrontEnd: Create Emergency Contact- Bug fix - 


htharker42:
when "Add New Phone Number" is selected to add an additional number, container height is not dynamically adjusted and buttons to submit phone number are pushed below the footer.

marcusjensen15:
Hey Hugh,

I believe I've fixed it. Please check out my code as this is my first pull request on the project aside from the README. My goal was to be as minimally invasive as possible when fixing this bug.

branch is called: emergency-contact-bug-fix

One thing I did that I'm not completely sure is kosher: I added a 'let' variable titled 'resize' at the top of addEmergencyContact.js which is accessible to all of the constants below. This was what I found to be the DRYest way to apply styling to all of the necessary fields on the "Add New Phone Number" click.

Thanks, and if this isn't up to standard please let me know what I can fix :)

Chris-Boe:
Fixed also by issue #231

PR #216, opened 2020-06-22
Emergency contact bug fix

marcusjensen15:
(Response to Issue #206)

Hey Hugh,

I believe I've fixed it. Please check out my code as this is my first pull request on the project aside from the README. My goal was to be as minimally invasive as possible when fixing this bug.

branch is called: emergency-contact-bug-fix

One thing I did that I'm not completely sure is kosher: I added a 'let' variable titled 'resize' at the top of addEmergencyContact.js which is accessible to all of the constants below. This was what I found to be the DRYest way to apply styling to all of the necessary fields on the "Add New Phone Number" click.

Thanks, and if this isn't up to standard please let me know what I can fix :)


Chris-Boe:
I can't seem to reproduce the bug so I'll let Hugh review and test this one. This is what I see:
image

marcusjensen15:
@Chris-Boe good question. I had the same issue initially :)

You need to begin filling out the fields. Once the first few are filled out the 'Add New Phone Number' option will appear.

In this branch, once 'Add New Phone Number' is clicked, the input fields, font size, and buttons will resize to fit within view and not overlap with the footer.

marcusjensen15:
Just added an additional scss rule: The input line height and font size will now only resize if the user is on a laptop screen. it is now congruent with the resize rules for the buttons on the page.

Chris-Boe:
Hey @marcusjensen15 , I updated a bunch of the stylings throughout the app in #267 that were messing with this page and others. This was one of the issues fixed by the general changes in that PR, mind if I close this one?

marcusjensen15:
Hey @Chris-Boe my bad! I was offline traveling for the past week. Yes, please close my branch out if it is conflicting :)

Issue #205, opened 2020-06-15
Frontend: Add New Property -- cancel button


htharker42:
Cancel Button should route back to /manage/properties

marcusjensen15:
Built a fix for this on branch:  issue205-cancel-button-route and created pull request. Let me know if there is anything I did incorrectly or need to fix!

marcusjensen15:
Merged in to dev, closing issue

PR #217, opened 2020-06-22
Issue205 cancel button route

marcusjensen15:
fix for issue: #205 . Please let me know if there is anything I need to fix.

### DIFF
--- a/src/views/addProperty.js
+++ b/src/views/addProperty.js
@@ -173,4 +173,4 @@ const formHandler = (data, context) => {
     }
 }
 
-export default AddProperty;
+export default withRouter(AddProperty);
--- a/src/views/addProperty.js
+++ b/src/views/addProperty.js
@@ -3,6 +3,8 @@ import { Form, Field, Formik } from 'formik';
 import * as Yup from 'yup';
 import UserContext from '../UserContext';
 import * as axios from 'axios';
+import { withRouter } from 'react-router-dom'
+
 
 const validationSchema = Yup.object().shape({
     name: Yup.string()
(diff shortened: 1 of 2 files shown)


src/views/addProperty.js
```diff
@@ -155,7 +157,7 @@ export class AddProperty extends Component {
                                         </div> */}
                                         <div className="container-footer">
                                             <button className={`${isValid && "active"} save_button button is-rounded`} type="submit" disabled={isSubmitting}>SAVE</button>
-                                            <button className="button is-dark is-rounded" onClick={()=>{console.log("cancel pressed")}}>CANCEL</button>
+                                            <button className="button is-dark is-rounded" onClick={()=>{this.props.history.push('/manage/properties')}}>CANCEL</button>
```
Chris-Boe: Because we're using React-Router, we could just use the <Link> component instead here https://reacttraining.com/react-router/web/api/Link

In fact, we should be doing this throughout the app where we have navigation buttons, but we're probably not very consistent yet 😄

Chris-Boe (changes requested):

src/views/addProperty.js
marcusjensen15: Totally, that is my bad. I'll revise it so it uses <Link> instead of the way I have it written. I agree, <Link> is much more straight forward, I'd just seen someone use the 'this.props.history.push' method somewhere else in the code :)

marcusjensen15:
Just revised the route so it leverages the <Link> react-router-dom component. Let me know if there is anything else that needs to be fixed!

Chris-Boe (approved):
Nice! Looks and works great!

Issue #203, opened 2020-06-15
FrontEnd: Add new staff member


htharker42:
Remove name prefill on Add New Staff member page. 
Add required field tags.

marcusjensen15:
Added 'required' tag within InfoField component (inside of addStaffMember.js).

PR #218, opened 2020-06-22
Response to issue: #203 add 'required' tag within the InfoField input component.

marcusjensen15:
Response to issue: #203 add 'required' tag within the InfoField input component.


Chris-Boe (approved):
The form hasn't actually been hooked up to the backend yet but this is a start 😄

PR #220, opened 2020-06-25
Table pagination

zfregin:
Check Pagination for Ticket List table

Run npm install to add the react-bootstrap-table-pagination package

You can use the backend branch ticket-pagination-seed-data to generate enough tickets for multiple pages in the table.


Chris-Boe (approved):
This looks great! We should definitely get this into a reusable grid so we can use it throughout the app!

PR #223, opened 2020-06-26
added Unit field to form render and Yup validation schema

### LINKED ISSUE #200: FrontEnd: Add Second Address line for Unit
Add Unit to Address line in front end. 
Line should be optional 
"unit": string

mswanson-me:
Placeholder text in the form is set as "Unit number", but I am not sure if that should be "Apartment/Unit number" or "Apartment or Unit number" or some variation.

### DIFF
--- a/src/views/addProperty.js
+++ b/src/views/addProperty.js
@@ -12,8 +12,6 @@ const validationSchema = Yup.object().shape({
     address: Yup.string()
         .max(250, "*Address can't be longer than 250 characters")
         .required("*Address is required"),
-    unit: Yup.string()
-        .max(50, "*Unit can't be longer than 50 characters"),
     city: Yup.string()
         .max(50, "*City can't be longer than 50 characters")
         .required("*City is required"),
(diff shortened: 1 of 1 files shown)


Chris-Boe (changes requested):
so according to the mockup (page 15 of https://drive.google.com/file/d/1YqboQogczYm1HkyRqEtVSzeQ61T9hWU2/view) it looks like "Units" is the number of Units that a property contains. The mockup also has it as the last line.

Let's make those changes to align with the mockup. Once this page gets hooked up to the API, we will also need to have "Units" added as a parameter to the Property resource on the backend

mswanson-me:
The requested changes have been made.

src/views/addProperty.js
```diff
@@ -132,6 +135,19 @@ export class AddProperty extends Component {
...
+                                                onChange={handleChange}
+                                                value={values.units}
+                                                placeholder="Number of units"
+                                                error={errors.units}
+                                            />
+                                            {errors.units ? (<div className="error-message">{errors.units}</div>) : null}
+                                        </div>
                                         {/* This field needs to be included after
                                             the units parameter is added to the api endpoint */}
                                         {/* <div className="form-row columns">
```
Chris-Boe: Looks like there's an older commented out version of this field right here so we can get rid of it 😄

Chris-Boe (changes requested):

src/views/addProperty.js
mswanson-me: LOL that's pretty funny. Fixing this now.

mswanson-me: Give it one more shot.

Chris-Boe (approved):
Nice job!

Issue #207, opened 2020-06-15
Frontend: Properties Search Feature 


htharker42:
Add search feature to box on */manage/properties

marcusjensen15:
Added this feature in branch #227 . User is able to search by property name, manager, and address. Let me know if there are any revisions I need to make!

PR #227, opened 2020-07-01
Issue207 properties search

marcusjensen15:
This branch is a resolution to issue #207 . The user is now able to search through all of the properties By: Name, Property Managers, and Address. The search looks for exact character match. 

Added 'Search' and 'Clear' buttons make UI better as well.

### DIFF
--- a/src/scss/pages/_properties.scss
+++ b/src/scss/pages/_properties.scss
@@ -19,6 +19,10 @@
               margin-top: 0px;
               margin-left: 15px;
             }
+            .clearButton{
+              background-color: #F5575A;
+            }
+
         }
 
         .button {
--- a/src/scss/pages/_properties.scss
+++ b/src/scss/pages/_properties.scss
@@ -14,6 +14,11 @@
             .search {
               width: 60%;
             }
+
+            .button {
+              margin-top: 0px;
+              margin-left: 15px;
+            }
         }
 
         .button {
--- a/src/views/properties.js
+++ b/src/views/properties.js
@@ -54,13 +54,15 @@ const selectRow = {
 export class Properties extends Component {
     constructor(props) {
         super(props);
-        
+
         this.state = {
         properties: [],
+        filteredProperties: [],
+        isFiltered: false
         }
 
         this.getProperties = this.getProperties.bind(this)
-    }    
+    }
 
     componentDidMount() {
         this.getProperties(this.context);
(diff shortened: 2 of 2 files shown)


Chris-Boe (changes requested):
Nice work so far! Most of the searching works great
A couple issues I'm concerned about though
* I'm only able to search via property Name - I don't get any results when using text from other columns
* Clicking "Search" with an empty search bar should just return everything, but instead it's filtering everything out so there aren't any results
* We should hook up the keyboard controls for this so that "Enter" activates the Search button

I think we could make the search be an as-you-type search instead, since that's what it looks like the mockups are implying, but we can bring this up to Hugh! If that's the case, we could eliminate bullets 2 and 3 above

marcusjensen15:
Thanks! It actually does let you search by property managers and address, it just looks for an exact char match so if you include a space at the end of the search it won't find what you are looking for. Is this an issue?

marcusjensen15:
bullet points one and two are now handled. Search now encompasses partial strings regardless of case, searching nothing, and users accidentally including spaces in the beginning/end of search query.

on the third bullet, we can let hugh make the call on search updating on each keypress.

marcusjensen15:
Just chatted with Hugh, he prefers buttons to keypress search. If everything looks good we should be ok.

Chris-Boe (changes requested):
Looking better! I'd still like to add the Enter button functionality so you can type, press enter, and it performs the search.
I did have this happen though - I searched for "3" and got a property that doesn't contain a "3" 🤔 
image

marcusjensen15:
Totally! I'll make it so it will run when either the button is clicked or when the 'enter' button is hit.

It is strange why it is returning that property that doesn't contain a '3'. I'll get to the bottom of it and fix it!

marcusjensen15:
Couple improvements to this: 

1. User can now only search by 'visible' property details. Before when you were searching '3' and it was returning properties that did not contain a '3', it is because the '3' was buried in the object as an ID. That doesn't happen anymore.

2. Search function now executes when a user clicks the 'search' button or presses the 'enter' key. 

If this looks good and you merge it into development, on a new branch I can modularize all of this out as a functional component to use in other areas of the app :)

Chris-Boe (approved):
Nice job, works great! Thanks for making those changes!

Issue #228, opened 2020-07-01
FrontEnd: Bug Fix - on /manage/properties, the table is overlapping into the footer. 


marcusjensen15:
Will set up scrolling property.

marcusjensen15:
Resolved in branch #229

PR #229, opened 2020-07-01
Issue208 footer overlap fix

marcusjensen15:
Fix to issue #228 add y-overflow scroll property to div so you can now see all of the results

### DIFF
--- a/src/scss/pages/_dashboard.scss
+++ b/src/scss/pages/_dashboard.scss
@@ -48,7 +48,7 @@
 	}
 	&__footer {
 		//MH - temp
-		position: fixed;
+		position: relative;
 		z-index: -100;
 		background-color: $lighterGray;
 		color: $blue;
--- a/src/scss/pages/_dashboard.scss
+++ b/src/scss/pages/_dashboard.scss
@@ -48,7 +48,7 @@
 	}
 	&__footer {
 		//MH - temp
-		position: relative;
+		position: fixed;
 		z-index: -100;
 		background-color: $lighterGray;
 		color: $blue;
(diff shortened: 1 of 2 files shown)


Chris-Boe (changes requested):
The scrolling and margin is great but the real issue with the footer is that it has an incorrect z-index property and anything else on the page hides it.

Let's change the z-index of the dashboard__footer class (which should also just be called "footer" and should be in it's own footer.scss class) to 50
Let's also change the z-index of the left nav's .is-sidebar-menu class to 100

Chris-Boe:
We also have every page re-implementing the scss class for it's "container" (add-property__container, properties__container, dashboard__container, etc all with the same css rules), which we could really just pull this out as a common class or even create a container div in App.js to wrap the Switch component and automatically apply styling and whatever else we want to the app's right side container

Chris-Boe:
Also a tip on git commits: feel free to do any amount of "rebasing" you want! If you have a commit that you want to undo, you can take it out, or if you want to edit the commit's message or changes, you can edit it for a cleaner PR 👍 
There are always multiple ways to accomplish essentially the same thing in git but as an example: If I was on the latest version of this branch I could just delete the last two commits like this
* `git rebase -i HEAD~2` which means open an interactive menu to change up the last two commits
* change the "pick" on both those commits to "drop"
* Exit the interactive window by clicking ESC then `:wq` to write and quit
* Once done rebasing, you update the online branch with a `git push -f`

OR you drop those commits by doing:
* `git reset --hard HEAD~2`
* then a `git push -f` to update the online branch

Chris-Boe (approved):

Chris-Boe:
Created a new issue for the footer and container stuff #231

Issue #195, opened 2020-06-11
Adjust user model to include properties for Property Managers


DaveTrost:
**Expected Behavior**
When GET /user/:id is called for a Property Manager, the properties they manage and the tenants that belong to those properties should be returned

pseudo-code from Hugh on how to accomplish this:
```
if role == "PROPERTY MANAGER":
 return user.json w/ properties
```

dbridenbeck:
For what its worth, below is the dummy data I'm currently using. If the returned data resembled this, it would be awesome. If it can't, I'm sure we can find a way on the front end to still make it work.

```
export const PROPERTY_MANAGER_DATA = [
	{
		id: "1",
		firstName: "Jim",
		lastName: "Oliver",
		fullName: "Jim Oliver",
		phone: "(503) 123-1234",
		lastUsage: "9/12/2019",
		status: "Active",
		email: "john.oliver@propertiespdx.org",
		properties: [
			{
				name: "Jim's Mansion",
				streetAddress: "1234 Burnside Street",
				city: "Portland",
				state: "OR",
				zip: "97211",
			},
			{
				name: "Meerkat Mansion",
				streetAddress: "4552 Burnside Street",
				city: "Portland",
				state: "OR",
				zip: "97231",
			},
		],
		tenants: [
			{
				name: "Brendan Smith",
				property: "The Magnolia",
				unit: "Unit #10",
				phone: "(503) 123-1234",
			},
			{
				name: "Maggie Taylor",
				property: "The Magnolia",
				unit: "Unit #2",
				phone: "(503) 123-1234",
			},
			{
				name: "Brock Hampton",
				property: "Gibon Gardens",
				unit: "Unit #8",
				phone: "(503) 123-1234",
			},
			{
				name: "Jason Nyguen",
				property: "Gibon Gardens",
				unit: "Unit #12",
				phone: "(503) 123-1234",
			},
		],
	},
	{
		id: "2",
		firstName: "Bob",
		lastName: "Smith",
		fullName: "Bob Smith",
		status: "Pending",
		phone: "(503) 123-1434",
		email: "bob.smith@propertiespdx.org",
		lastUsage: "9/12/2019",
		properties: [
			{
				name: "Bob's Mansion",
				streetAddress: "1224 Hill Street",
				city: "Portland",
				state: "OR",
				zip: "97777",
			},
			{
				name: "Orange Courts",
				streetAddress: "2222 Dover Street",
				city: "Portland",
				state: "OR",
[...]

Issue #183, opened 2020-05-21
FRONTEND: Create a form component for 'Add' pages


asaltveit:
As seen on pages 6, 7, 12, 15, 19, 23, etc. in https://drive.google.com/file/d/1YqboQogczYm1HkyRqEtVSzeQ61T9hWU2/view

PR #230, opened 2020-07-02
View property managers

### LINKED ISSUE #147: Create a View Property Manager page
Page 17 of the mockups https://drive.google.com/file/d/1YqboQogczYm1HkyRqEtVSzeQ61T9hWU2/view

dbridenbeck:
### Purpose 
- This resolves #147 
- Create views for /manage/managers and manage/managers/:id (see slides 16 and 17 on the Figma design)
- I also created /components/ToggleEditTable, a reusable component for handling tables that need to be editable when a Pen icon is clicked (I believe manage/properties/:id and manage/tenants/:id will also rely on this UI):

Screen Shot 2020-07-01 at 5 19 49 PM

### Validating
1. Run frontend and backend, login
2. Click "Property Managers" link in side navbar, should take you to a table with two property managers in it
3. Table's sort should work as expected, search bar was not implemented (I believe this is a separate ticket, #207?)
4. When property manager name is clicked, should take you to an individual manager's page
5. Clicking edit button on manager page should enable an editable form
6. Clicking submit will update the component's state and alert you with the saved changes (this alert is just a placeholder until the backend connects tenants and properties with property managers in #195)
7. While editing contact info, Pen icon should be disabled

### Background
- Note that currently these views are using dummy data. Once #195 is implemented we will be able to hook this up to the backend
- I haven't added tests for ToggleEditTable or the managers view, but can add them if we feel they are needed

### Questions
- In `<Manager />` I used useLocation instead of useParams simply because I wasn't able to successfully test the component when using useParams with Enzyme. Ultimately I switched to using react-testing-library for testing, so if we feel that we want useParams instead, I can take another stab at this!
- My creation of ToggleEditTable seems relevant to #183. Perhaps the form I created here would be helpful in that effort?
- What's the expected behavior for clicking the Invite button on /manage/managers ?
- I just realized pagination isn't working for /manage/managers :( should I add that to this PR or tackle that in a separate issue?
- I don't want to rock the boat, but have we considered using CSS modules to avoid naming conflicts? I tried mirroring my styling from other similar views, and noticed some views are using classes defined in another view's scss file, which could lead to some hair-pulling issues down the line. I'd be down to tackle this migration if it seems like a good idea, if not, no worries! :)


Chris-Boe (changes requested):
Looks like a bunch of stuff got edited in the /stashed folder, so let's kick all those changes out 😄

Chris-Boe:
It looks like there was a huge formatting commit done but the code actually added in this commit does not follow the same formatting standards - Do we want to keep the swath of formatting changes or discard those for now? I believe some of the formatting changes are actually against what the group decided on (not that it's a big deal as long as things are consistent) but it makes it a big hard to see what actually needs to be reviewed in this PR

Chris-Boe:
After chatting on Slack, PR #238 will replace this one, so once that PR is approved, we can close this one

Chris-Boe:
The other PR is looking really good so far, so I'm going to close this one!

Issue #246, opened 2020-07-16
Modualize search so it can be used in other areas of the app


marcusjensen15:

marcusjensen15:
Issue resolved in search-component branch and ready for review. Check out pull request #247

PR #247, opened 2020-07-16
Search component

marcusjensen15:
I built a component for Search that can be imported around the app wherever needed. Instructions on how to use it have been well documented within the comments of src/components/Search/index.js . Additionally, there is a working example of <Search/> being used in src/views/properties.js.

This is a fix for issue #246 

Here is what you need to make it work:

Data (See src/views/properties.js) (all should be state variables):

1. Input data (passed as input prop).

2. Where you want the data output (passed as outputLocation prop).

3. isFiltered (true/false) location (passed as isFilteredLocation prop).

4. The message you want displayed as a placeholder. (passed as placeholderMessage prop)

Functions: Need to write two functions in whatever parent component you import <Search> into.

1. Function to set 'isFiltered' state (T/F) in the parent component. Pass this function to <Search> via setIsFilteredStateFalse props. (See src/views/properties.js - setIsFilteredPropertiesFalse for example)

2. Function to set 'filteredResults' state within the parent component, and also 'isFiltered' state back to true. (See src/views/properties.js - setOutputState for example)

It works the exact same as the previously merged search in properties.js.

Thanks and let me know if you find any issues or anything that needs to be fixed!

### DIFF
--- a/src/components/Search/index.js
+++ b/src/components/Search/index.js
@@ -45,10 +45,10 @@ function Search(props) {
     return (
       <div className="search-section">
         <input className="input search is-rounded" id="searchQuery" onKeyDown={enterSearchHandler} placeholder="Search properties by name, address, or property manager"></input>
-          <button className="save_button button is-rounded" onClick={this.searchProperties}type="submit">
+          <button className="save_button button is-rounded" onClick={searchProperties}type="submit">
              Search
           </button>
-          <button className="save_button button is-rounded clearButton" onClick={this.clearSearch}type="submit">
+          <button className="save_button button is-rounded clearButton" onClick={clearSearch}type="submit">
              Clear Search
           </button>
       </div>
--- a/src/components/Search/index.js
+++ b/src/components/Search/index.js
@@ -7,7 +7,7 @@ import './search.scss';
 
 //1. Input data (passed as input prop).
 
-// 2. Where you want the data output (passed as outputLocation prop).
+//2. Where you want the data output (passed as outputLocation prop).
 
 //3. isFiltered (true/false) location (passed as isFilteredLocation prop).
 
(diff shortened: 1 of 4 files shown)


src/components/Search/index.js
```diff
@@ -0,0 +1,71 @@
+import React, {useState} from 'react';
+import './search.scss';
+
+//Necessary for <Search> to work:
```
Chris-Boe, line 4: Thanks for adding documentation!

src/components/Search/index.js
```diff
@@ -0,0 +1,71 @@
...
+          };
+      props.setOutputState(output, true)
+      }
+  };
+
+
+    return (
+      <div className="search-section">
+        <input className="input search is-rounded" id="searchQueryComponent" onKeyDown={enterSearchHandler} placeholder={placeholderMessage}></input>
+          <button className="save_button button is-rounded" onClick={searchProperties}type="submit">
```
Chris-Boe, line 62: We could probably use the Button component here from https://github.com/codeforpdx/dwellingly-app/blob/development/src/components/Button! Then we can get rid of the duplicate css

src/components/Search/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+    const searchProperties = () => {
+      let allData = input;
+      let output = [];
+      let searchQuery = document.getElementById("searchQueryComponent").value.toLowerCase().trim();
+
+       if(searchQuery.length > 0){
+          for (var i=0;i < allData.length; i++) {
+              let dataPoint = Object.values(allData[i]);
+              delete dataPoint.id;
```
Chris-Boe, line 48: Looks like we might be running into the `id` being a part of the search function again
image

Chris-Boe (changes requested):
Nice work! Love to see reusable components 😄

Chris-Boe (approved):
Nice job!

Issue #244, opened 2020-07-09
FRONTEND: Create Mobile Friendly Login/Create User Page


htharker42:
See [FIGMA Designs](https://drive.google.com/file/d/1wGLQI6d5-RK59E4P65wdotAXdNsSstRR/view?usp=sharing) for design guidelines. Under header LogIn Create Account Best if opened in PDF viewer. 

- Use existing routes if new routes mobile specific routes are needed use the /m/routename convention 
- For Existing pages use css [media breakpoints](https://www.w3schools.com/howto/howto_css_media_query_breakpoints.asp) to specify the different formats 

- [x] Create Login View

- [x] Create Account View

- [ ] Create Account Confirmation Alert 

- [x] Forgot Password View

- [x] Privacy Policy View

- [x] Terms and Conditions View 

If needed, a group can approach this topic. 

![image](https://user-images.githubusercontent.com/32914953/86998998-727f6d00-c166-11ea-82f0-2daf479b8cac.png)
![image](https://user-images.githubusercontent.com/32914953/86999043-86c36a00-c166-11ea-9177-9a3dd9eafc96.png)

Good night, and good luck.

marcusjensen15:
Tackled Login View in Pull Request #249 . Awaiting review. 

Also, I can make 'Login with Google' into a component if it would be helpful.

Chris-Boe:
Let's try to make pages as responsive as possible so there's no need for a separate "desktop" and "mobile" view. Our Header and main-container (wrapped around every route in App.js so it applies to everything) are already responsive, so for most of these we can just use the normal page with some small changes to the content's styling

Chris-Boe:
Closing this and breaking out the last piece as a new issue

PR #249, opened 2020-07-20
Mobile login view

marcusjensen15:
Response to 'create login view' task on issue #244 .

Login page now looks good across mobile devices. Was able to achieve this only using media query within _login.scss. Also added in the google "G" icon embedded in the 'login with google button'

Let me know your thoughts and if there is anything I need to revise :)


src/scss/pages/_login.scss
```diff
@@ -80,3 +87,29 @@
     margin-bottom: 10px;
   }
 }
+.googleIcon{
+  height: 25px;
+  float: left;
+}
+
+@media (max-width: $breakpointMobile){
```
Chris-Boe, line 95: I'm getting a `SassError: Undefined variable: "$breakpointMobile".` when trying to run this locally 🤔 
We should add this variable to /scss/variables.scss and define it according to the common breakpoints. There might also not be a catch-all breakpoint here either, so we might need to adjust after playing with the sizes

Chris-Boe (changes requested):

src/scss/pages/_login.scss
marcusjensen15: Interesting! I wasn't getting that error on my machine. I'll look into it.

I defined $breakpointMobile as 1200px, and had success in the past using this as the general 'mobile screen size'. There could easily be a better way!

marcusjensen15: @Chris-Boe that was my bad, I had originally defined $breakpointMobile within scss/variables.scss but forgot to commit. I just pushed up a couple more commits containing that global styling variable along with some minor changes to the 'Login With Google" button.  

I did a bit of research online before defining $breakpointMobile and 1200px ends up being a good rule of thumb. Should we need to change it in the future we always can :)

Chris-Boe (approved):
Looks great!

Issue #145, opened 2020-04-16
Add a user role Enum list


Chris-Boe:
We want to have User Role as an enum with the following items:
* Pending
* Tenant
* Property Manager
* JOIN Staff (this assumes there is a hash table on the backend that indicates which users are admins)

Workflow:
Users register on the "Create Account" page and then are added as a User with a "Pending" role. An Admin has to approve this "Request for Access" and assign a role to the user before they can use the application.

Chris-Boe:
@zfregin Let me know if you're planning on adding an enum on the frontend with that PR you've opened. I think we'll also need one on the backend so that the /roles endpoint works!

PR #251, opened 2020-07-23
Request Access Logic

### LINKED ISSUE #213: Finish "/request-access" page logic. Allow admins to assign roles.
Currently the "/request-access" route has has a form that allows admin users to assign a role to requested users. However, there is no submit button. Finish the logic here.

zfregin:
Issues #213 and #145 

See Backend branch request-access for according changes
Refactored Request Access component
Check for use of Role Enum and post to backend

Is a re-seeding of the database necessary? YES
Any new dependencies to install? No


src/views/requestAccess.js
```diff
@@ -1,146 +1,123 @@
...
+  const [roleObject, setRoleObject] = useState({});
+  const [selectionOptions, setSelectionOptions] = useState([]);
+  const [currentSelection, selectionHandler] = useState("");
+  const [fName, setFirstName] = useState("");
+  const [lName, setLastName] = useState("");
+  const [emailAddress, setEmail] = useState("");
+
+  useEffect(() => {
+    console.log(process.env.REACT_APP_API_URL)
+    axios.get(`/api/roles`)
```
Chris-Boe, line 53: I'm getting a 404 when testing this since /roles does not exist in the backend 
image

Let's comment this line out and put in some dummy data (which won't really be that different from the real data) until we get that API endpoint setup

src/views/requestAccess.js
```diff
@@ -1,146 +1,123 @@
...
+    .catch((error) => {
+      alert(error);
+      console.log(error);
+    });
+  }, []);
+
+  const grantAccess = (role, firstName, lastName, email, id) => {
+    // Get role index from roleObject
+    let roleID = roleObject[role];
+    axios.patch(`/api/user/${id}`, {
```
Chris-Boe, line 75: Since I didn't get any roles, the dropdown still said "Select Role", and clicking the Grant Access button gives me a 400 Bad Request since a role ID wasn't sent along with the request.
image

Let's put in a condition on the Grant Access button that disables it until a role is selected

src/views/requestAccess.js
```diff
@@ -1,146 +1,123 @@
...
+				<InfoField label={"First Name"} changeHandler={setFirstName} info={firstName} />
+				<InfoField label={"Last Name"} changeHandler={setLastName} info={lastName} />
 				{/* <InfoField label={"Phone"} info={data['phone']} /> */}
-				<InfoField label={"Email"} info={email} />
+				<InfoField label={"Email"} changeHandler={setEmail} info={email} />
 				<hr className="line" ></hr>
 				<div className="sub-title sub-title-padding"> ASSIGN ROLE </div>
-				<RoleDropDown />
+				<RoleDropDown selectionOptions={selectionOptions} selectionHandler={selectionHandler}/>
 				<div className="button-padding">
```
Chris-Boe: Let's make sure there's an issue created to add the "Property" section like in the mockup, or we can do that work with this PR! Up to you
image

src/views/requestAccess.js
```diff
@@ -1,146 +1,123 @@
...
+  useEffect(() => {
+    setFirstName(firstName);
+    setLastName(lastName);
+    setEmail(email);
+  }, []);
 
 	return (
 		<>
 			<div className="request-page">
 				<div className="page-title"> Request Access </div>
```
Chris-Boe: The mockup says "Request for Access" for the title of the page, I'm not sure if it makes that big of a difference but it might add clarity that the admins are looking at a list of requests to _manage_ and not a page to _make_ a request of their own 🤷‍♂️

Chris-Boe (changes requested):
Really clean code! Overall it looks pretty great, we're just missing the necessary functionality for it to be properly hooked up on the backend

src/views/requestAccess.js
```diff
@@ -1,144 +1,123 @@
...
+  return (
+    <div className="section-row">
+      <div className="select is-rounded">
+        <select
+          defaultValue='default'
+          onChange={e => props.selectionHandler(e.target.value)}>
+          <option value='default' disabled defaultValue>Select Role</option>
+          {
+            props.selectionOptions.map((role) => 
+            <option>{role}</option>
```
Chris-Boe: One other thing: When we do a .map or generate JSX content like this we need to add a "key" attribute to the `<option>` tag like `<option key={index}>` (you can access index from the map function as a parameter: .map( (role, index) => ...)

You can see the error it produces in the console if we don't do this

Chris-Boe (changes requested):

src/views/requestAccess.js
```diff
@@ -1,144 +1,123 @@
...
-			    </div>
-			</div>
-			<div>
-				{viewPropertyManagerOptions && <PropertyManagerOptions />}
-				{viewJoinStaffOptions && <JoinStaffOptions />}
-			</div>
-		</div>
-	);
+const RoleDropDown = (props) => {
+  console.log(props.selectionOptions);
```
Chris-Boe: We can get rid of these console logs

src/views/requestAccess.js
```diff
@@ -1,144 +1,123 @@
...
+
+  const [roleObject, setRoleObject] = useState({});
+  const [selectionOptions, setSelectionOptions] = useState([]);
+  const [currentSelection, selectionHandler] = useState("");
+  const [fName, setFirstName] = useState("");
+  const [lName, setLastName] = useState("");
+  const [emailAddress, setEmail] = useState("");
+
+  useEffect(() => {
+    console.log(process.env.REACT_APP_API_URL)
```
Chris-Boe: We can get rid of this log too

Chris-Boe (changes requested):
Nice work! I think this is good to go once we resolve the conflicts and take a couple logs out 👍

src/views/requestAccess.js
```diff
@@ -1,144 +1,120 @@
...
+    setFirstName(firstName);
+    setLastName(lastName);
+    setEmail(email);
+  }, []);
 
 	return (
 		<>
-			<div>
-				<div className="page-title"> Request Access </div>
+			<div className="request-page">
```
Chris-Boe: So I actually made a common main-container that applies these common styles to all pages, so we don't need this div or the className on it anymore!

Chris-Boe (changes requested):
Just one more change and this one's good! 😄

Chris-Boe (approved):
🎉

Issue #261, opened 2020-07-30
Frontend: Add phone to form on signup page


dbridenbeck:
**Purpose**
Add phone to new user registration. Complete only once #260 is merged so that phone is available on the UserModel.

**Proposed Solution**
Note, if #183 is merged by the time you begin working on this, make sure to use the reusable form component :)

- Add a new Field to the Formik for phone number
- Currently the signup page doesn't use Yup and validationSchema, which I'd recommend refactoring to use Yup.
- Assuming you've refactored to use Yup, make sure to add some validation in the validationSchema! Something like this should be good enough: 
```javascript
phone: Yup.string()
    .min(
      5,
      "*Number must contain at least 5 digits to be a valid phone/text number"
    )
    .max(20, "*Numbers can't be longer than 20 digits")
    .required("*a valid phone number is required")
```

**Acceptance Criteria**
Please list the requirements the implementation of this feature should meet.
- [x] Submitting an invalid phone number should prevent the user from submitting a form
- [x] Submitting a valid phone number should allow the user to submit the form

**Mockup Screenshot or Page Number**
Page 2 of the Figma docs needs to be updated to have "Phone" added as a field.
![Screen Shot 2020-08-13 at 3 54 38 PM](https://user-images.githubusercontent.com/49624436/90194993-52eaee00-dd7d-11ea-9d8d-ed99c95c91ac.png)

Page 55 of Figma docs shows the need for users to have a phone number when they register
![Screen Shot 2020-07-29 at 9 23 06 PM](https://user-images.githubusercontent.com/49624436/88880483-b1db2f80-d1e1-11ea-991b-76a9847cc501.png)

PR #271, opened 2020-08-17
Issue261: Added phone number to signup and requestAccess pages

jeremydavidson:
https://github.com/codeforpdx/dwellingly-app/issues/261

I added phone number to the signup page. I noticed it wasn't getting saved through /api/register, so I found and fixed a bug on the back end. The requirements on issue #261 stated that phone number is required, so I made it required on the back end API too. I noticed that phone number wasn't displayed on the requestAccess page, so I added it there too.

As requested on the issue, I refactored to use Yup for validation, also refactored to a functional component and added a unit test.

Any special requirements to test? Sign up is the main page in scope, but I added phone number to requestAccess page too.

### DIFF
--- a/src/views/signup.js
+++ b/src/views/signup.js
@@ -13,11 +13,7 @@ function isValidEmail(email) {
 
 export class SignupForm extends React.Component {
 
-  async signup(firstName, lastName, email, password, confirmPassword){
-    console.log(firstName)
-    console.log(lastName)
-    console.log(password)
-    console.log(confirmPassword)
+  async signup(firstName, lastName, email, phone, password, confirmPassword){
 
     if (password !== confirmPassword){
       alert("Passwords Don't Match");
@@ -28,6 +24,7 @@ export class SignupForm extends React.Component {
       firstName: firstName,
       lastName: lastName,
       email: email,
+      phone: phone,
       password: password,
       confirmPassword: confirmPassword
     })
@@ -66,6 +63,7 @@ export class SignupForm extends React.Component {
               firstName: "",
               lastName: "",
               email: "",
+              phone: "",
               password: "",
               confirmPassword: "", 
             }}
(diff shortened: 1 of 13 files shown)


jeremydavidson:
I added a fix for unrelated navigation menu test which was failing, also fixed several lint warnings. The only remaining warnings are on "jsx-a11y/anchor-is-valid" which calls for a more widespread refactoring to change anchor links to buttons.

src/views/signup.js
```diff
@@ -1,179 +1,179 @@
...
-      }
-    })
-    .catch((error) => {
-      if (error.response){
-        console.log(error.response.data.message)
-        alert(error.response.data.message)
+const SignupForm = ({ history }) => {
+  const signup = async (firstName, lastName, email, phone, password, confirmPassword) => {
+    let response;
+    try {
```
dbridenbeck, line 16: Nice! I think try/catch blocks are something that's needed in more of our components.

src/views/signup.test.js
```diff
@@ -0,0 +1,69 @@
...
+    jest.spyOn(axios, 'post').mockResolvedValue({});
+    jest.spyOn(window, 'alert').mockImplementation(() => {});
+    container = render(
+      <UserContext.Provider value={{ user: mockNotAuthenticatedUser }}>
+        <SignupForm history={mockHistory} />
+      </UserContext.Provider>
+    );
+  });
+
+  it("should display an error when phone not populated", async() => {
```
dbridenbeck, line 30: Really stellar testing coverage here, nice work! Since this specific test is essentially testing that Yup is working as expected, is this test actually needed? Open to feedback here, as I'm not a testing guru :)

dbridenbeck (approved):
Awesome work, Jeremy! Just had a question about a test in`signup.test.js` but other than that it looks great.

p.s. - Nice work getting rid of all those linter warnings and generally keeping things tidy :)

src/views/signup.test.js
jeremydavidson: To be pure, it would be best to mock out Yup. But, it does seem valuable to test whether an Error message displays or not. Another option might be to test that the API is NOT called as a result of this button press.

src/views/signup.js
jeremydavidson: For async operations, it can be a more readable alternative to Promise chain handling.

Chris-Boe (approved):
Nice work! Thanks @jeremydavidson it works great!

PR #285, opened 2020-08-27
Add settings page

### LINKED ISSUE #153: FRONTEND: Create the Settings page
Page 55 of the mockups https://drive.google.com/file/d/1YqboQogczYm1HkyRqEtVSzeQ61T9hWU2/view

dbridenbeck:
Closes #153 

Is a `npm install` necessary?: No
Any special requirements to test?: The form should submit successfully after editing it with new valid data. Note that I wrote some tests but had difficulty testing that `axios.patch` was called after the form is edited and submitted. There are three additional test cases I'd like to cover, but can't address until I fix this testing issue.

- [ ] Code has been tested and does not produce errors - Doesn't produce errors, but needs more tests.
- [ ] There isn't any unnecessary commented-out code - Will remove comments after I get help w/ testing :)


src/views/Settings/index.js
```diff
@@ -0,0 +1,134 @@
...
+});
+
+const formSubmitHandler = (context, data) => {
+  axios
+    .patch(`/api/user/${context.user.identity}`, data, {
+      headers: { Authorization: `Bearer ${context.user.accessJwt}` },
+    })
+    .then((response) => {
+      context.user.phone = response.data.phone;
+      context.user.email = response.data.email;
```
dbridenbeck: I get the feeling that directly modifying context like this isn't a best practice. The alternative seems to be to setup a Context Store and then useReducer() to issue actions that modify the store's state. This seems like a fair amount of work, and it seems like there's few places where we need to be modifying Context, so perhaps creating a Context Store could be it's own ticket if we decide to move in that direction.

Also open to other ideas/suggestions/solutions here :)

yhafez: Ok, so yes, we don't want to modify state directly like this because it'll mutate the state and when that happens, React doesn't register it as a change and so it fails to re-render which might be why you weren't seeing those values update. Instead, you'd want to use setState to replace the user state object with a new updated user state object. I'm kind of fuzzy on class-based React, but I'll do my best.

In App.js (like 152 in your PR), add setState to the context object. I'll name it "setUser":
` <UserContext.Provider value={{ user: { ...this.state.userSession }, setUser: this.setState, login: this.login, logout: this.logout }} >`

And replace lines 25 & 26 in this file (src/views/Settings/index.js) with something like:

`
const updatedUser = {...context.user, phone: response.data.phone, email: response.data.email};
context.setUser({ userSession: updatedUser });
`

So you'd use the spread operator on `context.user` to fill the new object (which I'm calling `updatedUser`) with the values from the user state before the update (`context.user`), and then replace the values for phone and email in `updatedUser` with the values submitted in the form (which I'm doing in the same line), and set the new object as the new value of `userSessions` using `context.setUser`.

The reason for this, as I understand it, is that if we manipulate the values for `context.user.phone` and `context.user.email` directly rather than using setState, we'd be mutating the state and React doesn't register that a change happened when we do that, so it doesn't re-render with the updated state we want to see and instead we still see the old state which might be why you were having trouble seeing the changes. 

We also need to use the spread operator here instead of doing something like this:

`
const updatedUser = context.user;

updatedUser.phone = response.data.phone;

updatedUser.email = response.data.email
`

because if we did it this way, `updatedUser` isn't creating a _new_ object, it's referencing the same object in memory that `context.user` references, so this would still result in us mutating the state rather than setting the state to a new updated state object using React's built-in "setState" method which is what we need to do for React to function properly. Whereas with the spread operator, we're creating a new object in memory which we then fill with the values from the old state before updating the phone and email values in the new object in memory. It sounds like it would use more resources, but that's the React way from my understanding, and I believe JavaScript's garbage collection is optimized to automatically free up the space of the old state object once we replace it and it no longer has any references pointing to it.

Hope that wasn't too convoluted and is helpful!

dbridenbeck: Hey Yahya!

Wow thanks for the in-depth dive on this! :) I'll look more into using context.setUser. Your comment re: using the spread operator makes total sense!

I don't actually think we'll need to use a class component, because I think we can use the useContext() hook to get access to the user's context and it's methods, but perhaps I'm missing something in your comment. Is there a specific reason you feel we'd want to use a class component vs. functional component? 

Also note that I'm using useEffect() with `context.user.phone, context.user.email` which forces a re-render on the component whenever those values change (my way of getting around the rendering issue I had earlier).

Thanks again Yahya! :)

yhafez: No problem! I definitely had to learn to use the spread operator instead of direct assignment situation the hard way and it was such a tough bug to work through! Using useEffect() like that is a really clever workaround, nice! My comment about class components was more so about trying to make sense of src/App.js, so there might be a bit of syntax change from my response when you export context.setUser (I think maybe you'd need to use `this` somewhere?).

yhafez:
[~] TODO: Update JWT stored in local storage as "dwellingly-access" with newly updated email and/or phone information; pending issue #288 

- I implemented a working solution, that could be optimized by resolving issue #288, but in the meantime, it's working for now!

dbridenbeck:
> * TODO: Update JWT stored in local storage as "dwellingly-access" with newly updated email and/or phone information; pending issue #288
> * TODO: Update form to integrate validation for email and phone fields

Re: Updating JWT we might

Chris-Boe:
Saw the TODO items so I put the "Under Construction" label on this PR. When you want it reviewed/merged later just take off that label! 👍

src/scss/pages/_settings.scss
```diff
@@ -0,0 +1,85 @@
+.add-property {
```
Chris-Boe: Let's update this to be for the Settings page (`.settings` instead of `.add-property`)and remove any unused styles we don't need

Chris-Boe (changes requested):
This looks great! Just wanted to have us cleanup the scss a bit.
Haven't looked too deep into the axios test but I'm okay with approving this even if it's not passing yet. We don't have many testing files in the front end yet and no requirements on them passing in the build so it's not a blocker in my opinion

Chris-Boe (approved):
Nice work!

PR #287, opened 2020-08-31
Signup/login google button

marcusjensen15:

This branch is in relation to issue #244 . This is a re-usable 'signup' or 'login with google' button that is formatted correctly for both desktop and mobile formats. We need this button in a few areas of the app, and I thought it would save time to just make a component out of it.

All of the documentation for this component is within: src/components/GoogleButton/index.js in comments at the top of the page.

This component is being actively used in src/views/login.js

Let me know if there is anything that needs to be fixed!

### DIFF
--- a/src/components/GoogleButton/googlebutton.scss
+++ b/src/components/GoogleButton/googlebutton.scss
@@ -1,0 +1,36 @@
+.login {
+
+  &__button{
+    color: white;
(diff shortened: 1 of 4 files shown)


Chris-Boe (approved):
Nice, clean work! This looks good

PR #291, opened 2020-09-03
Mobile create account

marcusjensen15:
I created one of the 7 views presented in issue #244 . This pull request is specifically for the signup page - mobile view.

This is really close to the wireframes. The only thing I changed really was:

- The cells for field input are a little bit smaller. If kept to the size on the wireframes the error fields would cause the buttons to run off the screen, and force the user to scroll down.

-In the rendering, each field had both a placeholder and a label. I felt like that was a bit redundant, but I'm happy to make that change if you like. 

I'm happy to make any changes always love your feedback!

### DIFF
--- a/src/components/GoogleButton/googlebutton.scss
+++ b/src/components/GoogleButton/googlebutton.scss
@@ -22,7 +22,4 @@
     width: 100%;
     font-size: 12px;
   }
-  .googleIcon{
-    margin-right: 3px;
-  }
 }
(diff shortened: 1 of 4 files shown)


Chris-Boe (changes requested):
Same comments from the privacy policy PR: let's cleanup some of the empty lines and make sure we are going for a more responsive look 
image

marcusjensen15:
Closing because this is a duplicate of the privacy policy branch #293 . They have a lot of the same commits.

PR #292, opened 2020-09-03
Use config file for issue templates

NickSchimek:
This allows us to select a project automatically for new issues.

Ensures new issues are automatically added to the kanban board.


Chris-Boe (approved):
Ah nice! I thought I had set that option in the project but I guess not! Good to know this is an option

Chris-Boe:
Tested after it was merged and it works! 👍

Issue #143, opened 2020-04-11
FRONTEND: Create Tenant information page


Chris-Boe:
Page 13 of the FIGMA mockups
https://drive.google.com/file/d/1YqboQogczYm1HkyRqEtVSzeQ61T9hWU2/view

PR #311, opened 2020-09-12
Issue143: Added Tenants view with search for JOIN staff.

### LINKED ISSUE #304: Connect the Manage Tenants page to the backend
**Purpose**
A working Manage Tenants page like on page 29 (also page 46 for more info) of the mockups

**Proposed Solution**

**Acceptance Criteria**
- [x] Get the list of tenants from the backend
- [x] Display the necessary info of the tenants in the grid
- [x] Provide support for Searching, Paging, Sorting, and Filtering, just like the Tickets page (try to use the same grid if possible! or the one from #235 if that is done)

*Leave out the UPDATE section above the table for now:*
image

Page 29
image

**Additional context**

jeremydavidson:
#143 

Is a `npm install` necessary? Yes
Any special requirements to test? Yes
- Added Authorization header to axios.defaults in App.jsx, so we don't have to add that header on each individual use of axios. Once we merge, we should let everyone know they no longer have to worry about the Authorization header.
- Uncommented and enabled tenants.js list to link to an individual tenant.
- Added Tenants view with search for JOIN staff.
- Added dependencies, will require `npm install`:
  - react-search-panel to match design for staff search
  - react-router and msw for unit testing.


src/views/Tenant/TenantView.js
```diff
@@ -0,0 +1,264 @@
...
+  /**
+   * Handle change in staff selections of search panel
+   * @param {*} selectedChoices
+   */
+  const handleChangeStaffSelections = (selectedChoices) => {
+    setStaffSelections(selectedChoices);
+  };
+
+  return (
+    <div className="manager__container">
```
Chris-Boe: I think we can remove this class on here. I added a main_container in App.js to add these styles to all pages instead of having every page rewrite these styles. Having both adds double margin/padding so it looks a little funky:
image

src/views/tenants.js
```diff
@@ -77,59 +82,71 @@ const selectRow = {
...
+                </Link>
+              </div>
+              <div className="search-section">
+                <input
+                  className="input search is-rounded"
+                  placeholder="Search Tenants by name, property, or JOIN staff"
+                ></input>
+              </div>
+              <div className="properties-list">
+                <BootstrapTable
```
Chris-Boe: We don't have to make these changes in this PR, but we should fix the table width and the columns to match the mockups on page 8:
image

src/views/tenants.js
```diff
@@ -77,59 +82,71 @@ const selectRow = {
...
-    //     })
-    // }
+  getTenants = (context) => {
+    axios
+      .get(`/api/tenants`, {
+        headers: { Authorization: `Bearer ${context.user.accessJwt}` },
+      })
+      .then((response) => {
+        const tenants = response.data.tenants;
+        tenants.forEach(tenant => {
```
Chris-Boe: a `.map` is probably better practice here but a forEach is fine

src/views/Tenant/TenantView.js
```diff
@@ -0,0 +1,264 @@
...
+            </h2>
+            <button
+              className={`rounded${isEditing ? "--is-editing" : ""}`}
+              onClick={handleEditToggle}
+              disabled={isEditing}
+            >
+              <i className="fas fa-pen icon" />
+            </button>
+          </div>
+          <div className="manager__contact">
```
Chris-Boe: If we are using styles from another page we might want to break them out into their own common styles section, or create a stylesheet for this TenantView page

src/views/Tenant/TenantView.js
```diff
@@ -0,0 +1,264 @@
...
+      });
+    }
+    return staffChoices;
+  };
+
+  /**
+   * Get a tenant
+   */
+  const getTenant = async () => {
+    const tenantResponse = await axios.get(`/api/tenants/${id}`);
```
Chris-Boe: I think the Promise oriented async style is easier to read and edit but if you have a strong feeling toward async await we don't have to change it 😄 This currently won't catch any errors or missing data that might happen from the API's

src/views/Tenant/TenantView.test.js
```diff
@@ -0,0 +1,56 @@
...
+  rest.get("/api/tenants/42", (req, res, ctx) => res(ctx.json(mockTenant))),
+  rest.get("/api/properties/mock1", (req, res, ctx) => res(ctx.json(mockProperty))),
+  rest.get("/api/tickets?tenant=42", (req, res, ctx) => res(ctx.json({}))),
+);
+
+beforeAll(() => server.listen());
+afterEach(() => server.resetHandlers());
+afterAll(() => server.close());
+
+describe("Tenant Component", () => {
```
Chris-Boe, line 38: Thanks for adding tests!

src/views/Tenant/TenantView.js
```diff
@@ -0,0 +1,264 @@
...
+    { id: "Closed", label: "Closed" },
+  ];
+  const [activeTab, setActiveTab] = useState(tabs[0].id);
+
+  /**
+   * Handle activating edit form
+   */
+  const handleEditToggle = () => setEditingStatus(!isEditing);
+  const onFormikSubmit = (values, { setSubmitting }) => {
+    setSubmitting(true);
```
Chris-Boe, line 65: Do we have an issue created to have the form actually save to the backend? If we don't do it with this PR

src/views/Tenant/TenantView.js
```diff
@@ -0,0 +1,264 @@
...
+      comp: <div />,
+    },
+  ];
+
+  /**
+   * When staff search input text changes, call API to find matching users.
+   */
+  useEffect(() => {
+    const loadStaff = async () => {
+      const staffResponse = await axios.post("/api/users/role", { userrole: 2, name: staffSearchText });
```
Chris-Boe: So Join Staff is userrole 3 while property managers is userrole 2

src/views/Tenant/TenantView.js
```diff
@@ -0,0 +1,264 @@
...
+              />
+            </div>
+          </div>
+
+          <div className="manager__contact">
+            <h1 className="secondary-title">JOIN STAFF</h1>
+          </div>
+
+          <div>
+            <SearchPanel
```
Chris-Boe, line 240: While I love server-side search tools, it might be easier for the user to see the whole list (or like the first 5 items) on click of the search component (without typing anything), since I don't think it will be likely to have hundreds or thousands of results for this search bar.

We should probably shorten the width of the Search component like the mockup.

Chris-Boe (changes requested):

src/views/tenants.js
jeremydavidson: Do you think the design implies a max-width on this table or is it 100% width with adequate padding?

src/views/tenants.js
jeremydavidson: I see we've got a fullName property in the tenants API now, so I'll remove this.

src/views/Tenant/TenantView.js
jeremydavidson: Fixed.

src/views/Tenant/TenantView.js
jeremydavidson: I prefer the async style because multi-level promise chains can be tough to understand. I should've added try/catch around each of these. But, I found that axios supports an error handling interceptor. So, I'll add one of those: see what you think.

src/views/Tenant/TenantView.js
jeremydavidson: I saw you added #304, so I'll work on that next. Thanks!

src/views/Tenant/TenantView.js
jeremydavidson: I'll fix that and add a constant for it, so this isn't a magic number.

src/views/Tenant/TenantView.js
jeremydavidson: I fixed the width and made an initial search call to load all staff. It will scroll when there are more results than would fill 200px height. If we aren't too concerned about this API call being too big or slow, I like the idea of letting the API do the filtering so you don't have to reload when data changes. But, we could code a client side onChange handler too. Would you prefer client side?

src/views/Tenant/TenantView.js
jeremydavidson: I changed it to 3, but it looks like we don't have any seed data for STAFF. So, this might be tough to test again without that. You can observe the API requests in developer tools, though.

src/views/tenants.js
jeremydavidson: I made it 100% for now.

src/views/Tenant/TenantView.js
jeremydavidson: Fixed

jeremydavidson:
Thanks @Chris-Boe for the feedback. I have addressed and commented on each one. I also did a merge from development, so hopefully there won't be any conflicts. This is ready for you to review again.

src/App.js
```diff
@@ -154,6 +158,13 @@ export class App extends React.Component {
       });
   }
 
+  /**
+   * Configure defaults for all axios requests in the App
+   */
+  updateAxiosDefaults = () => {
+    axios.defaults.headers.common['Authorization'] = `Bearer ${this.state.userSession.accessJwt}`;
```
Chris-Boe, line 165: Nice!

src/views/Tenant/TenantView.js
Chris-Boe: Serverside is good! I think the changes you made should be fine

Chris-Boe (changes requested):
Looks like something got a bit messed up with the Save/Cancel buttons below the form, they seem to be now appearing below/behind the bottom of the LeftNav menu 
image

jeremydavidson:
@Chris-Boe I fixed that issue with the save/cancel buttons: position is now relative, didn't see a need for it to be absolute either in the Property Managers page or here in Tenants.

Chris-Boe (approved):

PR #331, opened 2020-09-13
Update issue template config to auto add FE BE labels

NickSchimek:
This adds two new issue templates so that frontend and backend labels
can be applied accordingly.


Chris-Boe (approved):

Chris-Boe:
Thanks for updating this!

Issue #297, opened 2020-09-07
Documentation issue: Edit UI to meet accessibility standards (WIP)


Chris-Boe:
**Purpose**
What will this feature provide?
Meet Accessibility standards outline in #178 

**Proposed Solution**
Go through this doc and either fix them where possible or write up additional issues for sections that would require substantial work
https://www.figma.com/proto/4Tjz7YVGoiCEFX8emLS80E/Accessibility-Design-Audit?node-id=26%3A10&viewport=116%2C1179%2C0.25&scaling=min-zoom

**Acceptance Criteria**
Please list the requirements the implementation of this feature should meet.
- [ ]

**Additional context**
This issue is mostly a placeholder until I can take a deeper dive into the audit doc

PR #339, opened 2020-09-15
color overhaule, button overhaul,  aria labels added

SeHarlan:
#297 

In order to make the contrast better I had to over haul the color scheme. Some elements like the gray and black text still show up as insufficient contrast, but in oder to change that we would have to get rid of thier bulma class names, so for now I've just added shadows to help the contrast. 

In adding aria labels and testing buttons I noticed they needed some CSS love as well so i've gone in and changed a few parameters there to make them inline with the current color scheme and have more cohesive :hover and :active attributes. There are several different buttons from different components and with CSS classes. We may want to consider going through and making sure they are all the same so future CSS modification will be easier.

I'm still working on some of the auxiliary stuff from the audit but wanted to get this up now cause it's such a big potential design change.

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -15,7 +15,7 @@ html {
   min-height: calc(100vh - 150px);
   flex-direction: column;
   flex-grow: 1;
-  background-color: white;
+  background-color: $white;
   margin-bottom: 40px;
   padding: 25px;
 }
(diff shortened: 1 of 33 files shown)


beekman (changes requested):
Looks like there are some conflicts, otherwise looks good, just sort those out.

SeHarlan:
We can actually ax this, thank you!

PR #407, opened 2020-10-26
Assign property managers to AddProperty

### LINKED ISSUE #314: Add Assign Property Manager Component to Add New Property
**Purpose**
Add Assign Property Manager component to the Add New Property Page. 

**Proposed Solution**
Pull users where role = 2 
populate a dropdown with list of property managers.name
return the property manager as part of the property package 

**Acceptance Criteria**
- [ ]pulls list of users where role = 2
- [ ]populates dropdown 
- [ ]creates a property with assigned property manager

page 15
image

### LINKED ISSUE #369: Add Property Manager Drop down to Create Property
**Purpose**
Add a drop down which will list all the property managers in the database and allow the user to select one for the property. 

**Proposed Solution**
Use the /users endpoint and pass it 2 for property managers role this will return a list of the property managers in the system
populate the drop down menu with all usernames 
When a username is selected the user id is mapped to the propertymanager variable for the json package

**Acceptance Criteria**
- [ ]correctly pulls the propertymanager list and maps the names to the dropdown menus
- [ ]when a username is selected the userid is attached to the json object for the property.propertymanager
- [ ]server returns 201 

image

mtb116:
Closes #314 
Closes #369  

Created a new branch and copy/pasted changes from AssignPropertyManager because of really messed up commit history. Nick S gave me a walk through on rebase and his workflow to avoid that next time.

Assigned managers are saved by their id in an array to be sent off to the backend. 

This line I added at 199 is odd. I used library react-select just below but it would break having a value attribute. I added this Field, that is hidden, so it would work with Formik as I understood it.
```
<Field style={{ display: 'none' }} name="managers" value={this.state.assignedPropertyManagers} />
```

- [yes ] Code builds successfully
- [ does not] Code has been tested and does not produce errors
- [ I hope so!] Code is readable and formatted
- [ none] There isn't any unnecessary commented-out code
- [-] There aren't any unnecessary commits or files changed (shouldn't touch the /stashed folder unless the issue requests it)


Pipfile
```diff
@@ -0,0 +1,11 @@
...
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
+
+[requires]
+python_version = "3.8"
```
NickSchimek: Looks like `pipenv install` was ran on the front-end repo. 
This file can be deleted - It does not belong on the FE repo.

mtb116: Removed that file

src/views/addProperty/addProperty.js
```diff
@@ -1,10 +1,13 @@
...
-import { Form, Field, Formik } from 'formik';
+import { Form, Field, Formik, isEmptyArray } from 'formik';
 import * as Yup from 'yup';
-import UserContext from '../../UserContext';
+import UserContext from '../UserContext';
 import * as axios from 'axios';
 import { Link } from 'react-router-dom';
-
+import Select from 'react-select';
+// import { doPasswordReset } from '../../stashed/src/firebase/auth';
```
Chris-Boe, line 8: These can be removed

Chris-Boe (changes requested):
If we want to pull in a new package, we'll need to add it to the package.json list for it to be included in the node_modules. But we already have the `react-search-panel` included and that is being used for the same type of dropdown on the TenantView page so let's convert this page to using that, and remove the references to `react-select`. We can also uncommit the package-lock.json changes

PR #422, opened 2020-11-11
Change password payload

### LINKED ISSUE #418: Update payload for change password form
**Purpose**

Currently the change password form is only sending the 'password' field as a payload. It should send the entire form to the backend for processing so that the backend can do its own validation on the form. This will enable us to fix a bug that currently allows a user to change their password without inputting the correct current password.

**Acceptance Criteria**

- [ ] The payload request should include all fields

Jam52:
Payload changed to:
current_password
new_password
confirm_password

fixes #418

### DIFF
--- a/src/views/Settings/changePassword.js
+++ b/src/views/Settings/changePassword.js
@@ -24,7 +24,9 @@ const ChangePassword = () => {
     // TODO store and send only hashed password
     axios
       .patch(`/api/user/${context.user.identity}`, {
-        password: data.new
+        current_password: data.current,
+        new_password: data.new,
+        confirm_password: data.confirm
       }, {
         headers: { Authorization: `Bearer ${context.user.accessJwt}` },
       })


src/views/Settings/changePassword.js
```diff
@@ -23,17 +21,21 @@ const ChangePassword = () => {
...
-      .patch(`/api/user/${context.user.identity}`, {
-        password: data.new
-      }, {
-        headers: { Authorization: `Bearer ${context.user.accessJwt}` },
-      })
+      .patch(
+        `/api/user/${context.user.identity}`,
+        {
+          password: data.new_password,
+        },
```
NickSchimek: This is the payload that is being sent to the backend. 
Currently only the new_password is being sent. Please update to send all the form fields to the backend.

NickSchimek (changes requested):
Thanks for looking into this. The payload needs to be updated to send all the form fields to the backend. I highlighted in the below comment where this can be done.

Jam52:
Thanks Nick, I was concentrating too much on what the Formik elements were!

NickSchimek (approved):
Thanks @Jam52 - There were a bunch of style changes not related to this issue so I went ahead and removed them. I'm not sure what the style guidelines are for this repository, or even if there are any. However, I would prefer to only make the changes that are necessary. I also reverted the name changes, but kept the new payload params. Let me know if there are any questions. Thank you.

NickSchimek:
Verified manually. The request now sends the full payload. 

Screen Shot 2020-11-13 at 8 26 36 PM

@Chris-Boe We have a PR in the backend that is pretty much ready to go if this looks good. If your OK with these changes then I will merge both this and backend one at the same time when they're ready to go.

Jam52:
Thanks @NickSchimek I will ensure to check for a style guide and turn off / modify my extensions in future.

Issue #419, opened 2020-11-08
Update Tenant Info page lease option with calendar widget


Chris-Boe:
**Purpose**
What will this feature provide?
Allows the user to edit the lease dates on the Tenant information page.

**Proposed Solution**
If you have an idea on how to implement this feature, please share!
Add the Calendar Modal functionality (from the Add Tenant page) to the Lease option on the Manage Tenant page

**Acceptance Criteria**
Please list the requirements the implementation of this feature should meet.
- [ ] Should update the following two fields in the backend when the tenant gets saved:
    dateTimeStart = db.Column(db.DateTime, nullable=False)
    dateTimeEnd = db.Column(db.DateTime, nullable=False)

**Mockup Screenshot or Page Number**
Pages 11 and 30
![image](https://user-images.githubusercontent.com/8600792/98458757-c6b45100-2148-11eb-892e-6eefa89ec54d.png)
![image](https://user-images.githubusercontent.com/8600792/98458760-cc119b80-2148-11eb-8d1a-be6501d9087f.png)

**Additional context**

DaveTrost:
This is closed by https://github.com/codeforpdx/dwellingly-app/pull/456

PR #431, opened 2020-11-17
CalendarModal

SeHarlan:
#419
Reusable Calendar Modal that supplies a date range input.

!Important Note! 
In TenantView and AddTenant there is state initialized for the Date ranges titled <dateRange> in the form of an array with two date objects [startDate, endDate] but it is not hooked up to any service functions. I'm waiting on the backend and could probably use some help with where to plug that in.

Because these icons will be used "in" tables that are already made I had to do some kinda hacky CSS to get it positioned correctly see number 3 below (iconYposition) or dm Scott Harlan for more details.

### **Name**: CalendarModal
* **File path**: src/components/CalendarModal/CalendarModal
* **Input/Props**: title, dateRange, setDateRange, iconYPosition
* **Currently in use**: TenantView, AddTenant
* **Will be used in**: Anywhere a two date range is needed
* **Additional info**: 
  1. "dateRange" is [startDate, endDate] (as Date objects) 
  2. "setDateRange" is a setState Func
  3. Calendar icon is set to float:right and may need adjusting on the Y axis. Use iconYPosition to adjust this, preferably with "rem" so it will adjust with text sizing. e.i "-3rem" to shift it up or "4rem" to shift down.
  4. Initialize with either,
    ```
    const [dateRange, setDateRange] = useState([new Date(), new Date()]) 
    ```
  or the custom hook that has been supplied for the state.
    ```
    import CalendarModal, {useCalendarState} from '<filepath>/CalendarModal/CalendarModal'
    const [dateRange, setDateRange] = useCalendarState()


Is a `npm install` necessary? 
!!YES!! we must npm i to get the react-calendar library

Nothing I can think of

### DIFF
--- a/Reusable_Components_Doc.md
+++ b/Reusable_Components_Doc.md
@@ -146,3 +146,22 @@ _____
 2.  Toast is a function which takes two arguments. The first is the string you would like displayed in the notification. If no text is provided, you will see an error notice in the browser console. The second argument is the type of notification you would like displayed, which dictates the color of the notification. The options for this are "success" (green), "error" (red), "warn" or "warning" (yellow), "info" or "default" (blue). If no second argument is passed, the notification will default to "info" and render a blue toast notification. An example of this could look like `Toast("Property Added!", "success");`
 
 ---
+
+### **Name**: CalendarModal
+* **File path**: src/components/CalendarModal/CalendarModal
+* **Input/Props**: dateRange, setDateRange
+* **Currently in use**: TenantView, AddTenant
+* **Will be used in**: Anywhere a two date range is needed
+* **Additional info**: 
+  1. "dateRange" is [startDate, endDate] (as Date objects) 
+  2. "setDateRange" is a setState Func
+  3. To initialize with either,
+    ```
+    const [dateRange, setDateRange] = useState([new Date(), new Date()]) 
+    ```
+  or the custom hook that has been supplied for the state.
+    ```
+    import CalendarModal, {useCalendarState} from '<filepath>/CalendarModal/CalendarModal'
+    const [dateRange, setDateRange] = useCalendarState()
+    ```
+_____
--- a/Reusable_Components_Doc.md
+++ b/Reusable_Components_Doc.md
@@ -149,20 +149,20 @@ _____
 
 ### **Name**: CalendarModal
 * **File path**: src/components/CalendarModal/CalendarModal
-* **Input/Props**: title, dateRange, setDateRange, iconYPosition
+* **Input/Props**: title, calendarState, iconYPosition
 * **Currently in use**: TenantView, AddTenant
 * **Will be used in**: Anywhere a two date range is needed
 * **Additional info**: 
-  1. "dateRange" is [startDate, endDate] (as Date objects) 
-  2. "setDateRange" is a setState Func
+  1. "calendarState" is structured as follows:
+    ```
+    {startDate, endDate, setStart, setEnd} = calendarState
+    ```
+    Only startDate and endDate should ever be accesed (and then read only) the setState funcs are passed automatically to the calendars.
+  2. "calendarState" comes from the custom hook provided in the component file.
   3. Calendar icon is set to float:right and may need adjusting on the Y axis. Use iconYPosition to adjust this preferably with rem so it will adjust with text sizing. e.i "-3rem" to shift it up or "4rem" to shift down.
-  4. Initialize with either,
-    ```
-    const [dateRange, setDateRange] = useState([new Date(), new Date()]) 
-    ```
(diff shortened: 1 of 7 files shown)


NickSchimek:
@SeHarlan I'm unable to use the calendar to select dates. The calendar pops up, and I can select a date range, after that I cannot get the selected dates to populate the date field. Also is there a way we can use the calendar design shown in the Figma docs? I find the one calendar view for a date range rather confusing when the range isn't less than a month. The design shows two calendars popping up which probably isn't ideal, but I think would be more intuitive.

@htharker42 @Chris-Boe @DaveTrost Probably not related to this PR. I see the lease field is required. **It shouldn't be required** when creating a tenant as there can be unhoused tenants in the system that have not signed a lease yet. This made me realize there is not a crud view for creating leases - perhaps there should be? - I guess that is something we should probably bring up with our partner.

SeHarlan:
@NickSchimek I've got two calendars up now, I agree its much more intuitive even if its a little clunkier. The dates are kept in state but are currently not being sent anywhere until we decide how we want to handle the dates with form error handling and such.

SeHarlan:
I did kind of a weird workaround for the submit on addTenant but I think it would work. Although I've never used Formik so I'm not sure.
I tried to give it a crack on the TenantView page but that page is using a toggleEditTable component that I have no idea how to integrate the calendar state into it cohesively. I'd love to tackle this with anyone familiar with that component!

That might be a separate ticket though?? Not sure though, just let me know where you need me!

Chris-Boe (changes requested):
Got some merge conflicts after another recent PR

src/views/AddTenant/addTenant.js
```diff
@@ -40,6 +41,7 @@ export const AddTenant = () => {
   const [propertyOptions, setPropertyOptions] = useState([]);
```
Chris-Boe, line 41: Can you remove the lease object's .required in the validation schema on line 28? That will allow us to have unhoused tenants in the system which we'll need.

Reusable_Components_Doc.md
```diff
@@ -146,3 +146,23 @@ _____
 2.  Toast is a function which takes two arguments. The first is the string you would like displayed in the notification. If no text is provided, you will see an error notice in the browser console. The second argument is the type of notification you would like displayed, which dictates the color of the notification. The options for this are "success" (green), "error" (red), "warn" or "warning" (yellow), "info" or "default" (blue). If no second argument is passed, the notification will default to "info" and render a blue toast notification. An example of this could look like `Toast("Property Added!", "success");`
 
 ---
+
+### **Name**: CalendarModal
```
Chris-Boe, line 150: Thanks for adding this documentation!

src/views/AddTenant/addTenant.js
Chris-Boe: Let's also remove the Occupants .required property

src/components/CalendarModal/CalendarModal.jsx
```diff
@@ -0,0 +1,54 @@
...
+import Calendar from "react-calendar";
+import React, { useState } from 'react';
+
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faCalendarAlt } from "@fortawesome/free-solid-svg-icons";
+import './calendarModal.scss'
+import Modal from '../Modal';
+import { useEffect } from 'react';
+
+export default function CalendarModal({ calendarState, title = "Date Range", iconYPosition = "" }) {
```
Chris-Boe, line 11: The Calendar modal is looking great! There's a couple things I'd like us to add and one thing that seems like a bug I noticed:
* Can we add Cancel and OK buttons at the bottom right of the modal below the two calendars so the user can have a definite path to accept or reset the dates instead of clicking the X in the top right? The OK button would set the dates on click while the Cancel button would reset the dates back to the original dates they had when the modal opened
* We'll need to add some verification that the end date is *after* the start date and prevent the user from selecting an invalid date in the calendar
* There's apparently a hidden section below the calendar that changes the year when clicked:
datepicker

Chris-Boe (changes requested):

src/components/CalendarModal/CalendarModal.jsx
Chris-Boe: A nice to have: highlighting all the days shown between the start and end dates like this image
image

Chris-Boe:
As far as sending the dates to the backend, this is what the frontend is sending:
image

The backend is not doing anything with it now, so once we fix the above issues I think this PR is ready to go and we can handle the frontend and backend lease communication in a new issue

NickSchimek:
I created issue #452 to handle this request.

Chris-Boe:
Replaced by PR #456

PR #456, opened 2020-12-03
Calendar modal

SeHarlan:
This PR is an extension of PR #431 

#419
Reusable Calendar Modal that supplies a date range input.

### **Name**: CalendarModal
* **File path**: src/components/CalendarModal/CalendarModal
* **Input/Props**: calendarState, title (default is "Date Range"), resetOnClose (default is true), iconYPosition (optional),
* **Currently in use**: TenantView, AddTenant
* **Will be used in**: Anywhere a two date range is needed
* **Additional info**: 
  1. "calendarState" is structured as follows:
    ```
    {dateTimeStart, dateTimeEnd, setStart, setEnd} = calendarState
    ```
    Only startDate and endDate should ever be accesed (and then read only) the setState funcs are passed automatically to the calendars.
  2. "calendarState" comes from the custom hook provided in the component file. You can provide this hook start and end dates, otherwise it will default to both dates as the current date.
  3. Calendar icon is set to float:right and may need adjusting on the Y axis. Use iconYPosition to adjust this preferably with rem so it will adjust with text sizing. e.i "-3rem" to shift it up or "4rem" to shift down.
  4. The calendar state you have access to is only set once you click confirm, if you close the modal by default the state resets to the inital values, you can change this so that the temporary state stays the same even when closed
  5. Initialize with 
    ```
    import CalendarModal, {useCalendarState} from '<filepath>/CalendarModal/CalendarModal'
    const calendarState = useCalendarState()
    ....
    <CalendarModal title="Lease Range" calendarState={calendarState} resetOnClose={false} iconYPosition="2rem">

    ```

Is a npm install necessary?
Yes, we must npm i to get the react-calendar library

Nothing I can think of

### DIFF
--- a/Reusable_Components_Doc.md
+++ b/Reusable_Components_Doc.md
@@ -154,7 +154,7 @@ _____
 * **Additional info**: 
   1. "calendarState" is structured as follows:
     ```
-    {startDate, endDate, setStart, setEnd} = calendarState
+    {dateTimeStart, dateTimeEnd, setStart, setEnd} = calendarState
     ```
     Only startDate and endDate should ever be accesed (and then read only) the setState funcs are passed automatically to the calendars.
   2. "calendarState" comes from the custom hook provided in the component file. You can provide this hook start and end dates, otherwise it will default to both dates as the current date.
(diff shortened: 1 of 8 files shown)


Chris-Boe (approved):
Nice job!

Issue #435, opened 2020-11-18
Dashboard Links


DaveTrost:
**Purpose**
Link the cards on the dashboard to the detail pages that have been built. 

**Proposed Solution**
There are three cards on the dashboard. Two of these already have small ">" links. These links should go to the appropriate pages (currently, they link back to the dashboard) and the third needs a link added - this appears to be an oversight on the original Figma. Additionally, I think users will expect the entire title line of the card to be clickable as part of the link. And, I wouldn't argue if you made an argument for the entire card to be a clickable link.

**Please make sure you've attempted to meet the following coding standards**
- [ ] Code builds successfully
- [ ] Code has been tested and does not produce errors
- [ ] Code is readable and formatted
- [ ] There isn't any unnecessary commented-out code
- [ ] There aren't any unnecessary commits or files changed (shouldn't touch the /stashed folder unless the issue requests it)

**Mockup Screenshot or Page Number**
Page 4 of the [FIGMA designs](https://drive.google.com/file/d/1YqboQogczYm1HkyRqEtVSzeQ61T9hWU2/view)

NickSchimek:
@pogorobot is working on this.

PR #462, opened 2020-12-04
435 dashboard links

pogorobot:
The cards on the dashboard which are intended to link to their respective detail pages, are now clickable across the entire title line. (See issue #435) An additional backend pull request will provide the actual links to those pages, since the current structure accepts the link data as part of the API payload.

In addition, while testing this I discovered that many detail pages were *only* accessible via the nav menu - my new links, and simply typing the appropriate URL, would redirect to the dashboard. This turned out to be caused by the authorization logic setting its state twice, first in the app constructor and then again in componentDidMount. This pull request moves the authorization logic into the constructor, fixing the issue and roughly halving the app's load time as a side effect.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -44,16 +44,33 @@ export class App extends React.Component {
   constructor(props) {
     super(props);
 
-    this.state = {
-      userSession: {
-        isAuthenticated: false,
-        accessJwt: '',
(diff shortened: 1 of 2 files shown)


Chris-Boe (approved):
Looks good!

PR #483, opened 2020-12-17
482 - /manager/{id} now renders instead of crashes

### LINKED ISSUE #482: /manage/manager/{id} fails to render due to calendarState prop in toggleEditTable
**Describe the bug**
Attempting to view an individual property manager page produces a TypeError and does not render the view.

**To Reproduce**
1. Go to the dwellingly homepage
2. Click on 'property managers' in the left navbar
3. Click on a property manager name to view the individual property manager page

**Expected behavior**
Individual property manager pages should load successfully.

 - OS: mac catalina
 - Browser: chrome
 - Version: 87.0.4280.88

**Additional context**
I think this can be solved by editing toggleEditTable to have some conditional logic that only attempts to destructure the calendarState variables if calendarState is truthy

ccbass:
Closes #482 

Adds self to contributors.

Is a `npm install` necessary?  No.
Any special requirements to test? Click on specific Manager in Manage > Property Managers page.

### DIFF
--- a/src/views/Manager/index.js
+++ b/src/views/Manager/index.js
@@ -6,6 +6,7 @@ import * as axios from "axios";
 import UserContext from '../../UserContext';
 import TitleAndPen, { useEditingStatus } from "../../components/TitleAndPen";
 import Toast from '../../utils/toast';
+import { useCalendarState } from "../../components/CalendarModal/CalendarModal";
 
 import './manager.scss';
 
@@ -55,6 +56,7 @@ const Manager = () => {
   }, []);
 
   const { isEditing, setEditingStatus } = useEditingStatus()
+  const calendarState = useCalendarState()
 
   const tableData = managerData && [
     {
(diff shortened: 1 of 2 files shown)


src/views/Manager/index.js
```diff
@@ -132,6 +134,7 @@ const updateManager = (payload) => {
             isEditing={isEditing}
             submitHandler={onFormikSubmit}
             cancelHandler={onCancelClick}
+            calendarState={calendarState}
```
dbridenbeck, line 137: I don't know if the individual manager page actually needs calendarState, does it? Ultimately I think that whatever gets this page working is a good idea, but it might be confusing for other folks in the future if my hunch is right and this isn't necessary

ccbass: You are correct, Manager does not specifically use it.  However, might be useful in the future, in case something like tenure or hiring dates, etc wanted to be added.

Though, it might be better to pass the same argument that TenantView has.

ccbass:
@DaveTrost and I discussed this today and I found a different way to make this work without passing unnecessary data to the Manager component.  

Updated PR to follow: #486

PR #464, opened 2020-12-06
Backend to property managers

dbridenbeck:
### What issue is this solving?
closes #401 

### Any helpful knowledge/context for the reviewer?
1. I had to resolve merge conflicts over two occasions on this PR because life kept getting in the way for me to finish this PR, so a double check that those were properly merged would probably be wise.

2. Note how I implemented convertManagersDataForTable() on managers.js and how I'm rendering manager.status for the table. Not sure if that's really how we're hoping to implement this, but I gave it my best shot :)

3. Also note that my old laptop broke, I'm on a new one, and having trouble getting the backend up and running, so I'm unable to do any further testing until then, my apologies for the dumpster fire PR here :(

### Please make sure you've attempted to meet the following coding standards
- [x] Code builds successfully
- [x] Code has been tested and does not produce errors
- [x] Code is readable and formatted
- [x] There isn't any unnecessary commented-out code
- [x] There aren't any unnecessary commits or files changed (shouldn't touch the /stashed folder unless the issue requests it)

PR #484, opened 2020-12-18
Issue402 rename files

### LINKED ISSUE #402: Rename Files with JSX to *.jsx
**Describe the bug**
A lot of files which have JSX in them are labeled JS and not JSX. Please rename the file extensions correctly.

EstherGodoy:
- All component & view js files have been updated.
- I wasn't sure if we should .jsx the test files as well, I believe this would involve a small refactor to how jest is set up - which potentially could be a separate ticket?

**Additional notes added by @DaveTrost**

closes #402 

Any special requirements to test? please make sure to stop npm and re-run "npm start" as file imports need to be refreshed


dbridenbeck:
Woo hoo Esther, gettin' in that first PR! :) 

This looks good, but wondering if enums.js and constants.js need to be converted to .jsx, since they are straight up JS files with no JSX included?

Also, I *think* it's okay to keep the jest files with the `.js` extension. The docs for CRA suggest that .js is fine (see here https://create-react-app.dev/docs/running-tests/), but perhaps there's advantages of using .jsx instead.

Also for future PR's if you include the phrase "Closes #XYZ-ISSUE-NUMBER" it will auto-link to that issue, and auto-close that issue when this PR is approved and merged into development. Here's a link to an example of that in case what I'm saying doesn't make sense!: https://github.com/codeforpdx/dwellingly-app/pull/464

EstherGodoy:
I thought the same thing too @dbridenbeck - wasn't sure if we should just flood every single file for uniformity or if we wanted to retain data/obj only files as .js.

Sames goes for the test files.

@Chris-Boe let us know your thoughts.

Chris-Boe (changes requested):
Darren's right, so we should only update files to have a .jsx extension if those files have JSX in them

DaveTrost (commented):
Hi @EstherGodoy - Sorry it's taken a long time for me to review this PR. I agree with the changes requested above. 

Looks like the files that maybe need to be renamed back to (just) ".js" are the following:
- components/DashboardModule/data.jsx
- components/JoinStaffCard/data.jsx
- constants/index.jsx
- Enums/RoleEnum.jsx
- utils/*.jsx
- components/Tenant/index.jsx

DaveTrost:
I'm going to pick this one up from @EstherGodoy. There are only very minor changes left since Esther's work. Thanks Esther!

htharker42 (approved):
Looks good, thanks all!

PR #494, opened 2020-12-21
Add tenant validation to match (current) backend requirements

DaveTrost:
No issue in Git. Some errors were noted during new tenant submission by @NickSchimek, @aedwardg and @DaveTrost in a Slack convo. This PR fixes those errors

Any special requirements to test? Submit a new tenant, try to submit with/without a selected property and with/without lease dates. Note: because of the default values set by calendar modal, it is possible to submit a tenant without specifying lease dates, but at least the error text now directs users that the lease dates are required.

### DIFF
--- a/src/views/AddTenant/addTenant.js
+++ b/src/views/AddTenant/addTenant.js
@@ -43,7 +43,7 @@ export const AddTenant = () => {
   const [propertyErrorText, setPropertyErrorText] = useState("");
 
   const calendarState = useCalendarState();
-  const { dateTimeStart, dateTimeEnd } = calendarState;
+  const { dateTimeStart, dateTimeEnd, resetDates } = calendarState;
 
   useMountEffect(() => getProperties());
 
@@ -191,6 +191,7 @@ export const AddTenant = () => {
             setStaffSearchText("");
             setStaffSelections([]);
             setPropertyErrorText("");
+            resetDates();
             setIsValidationActive(false);
             setSubmitting(false);
           }}
@@ -314,7 +315,7 @@ export const AddTenant = () => {
                       Create New Property
                     </button>
                   </div>
-                  <h1 className="section-title">UNIT</h1>
+                  <h1 className="section-title">LEASE</h1>
                   <div className="form-row form-first-row">
                     <label
                       className="column is-one-fifth"
@@ -329,7 +330,7 @@ export const AddTenant = () => {
                       name="unitNum"
                       onChange={handleChange}
                       value={values.unitNum}
-                      placeholder="Unit Number"
+                      placeholder="Unit Number (Optional)"
                     />
                     {errors.unitNum ? (
                       <div className="error-message">{errors.unitNum}</div>
@@ -349,7 +350,7 @@ export const AddTenant = () => {
                       name="occupants"
                       onChange={handleChange}
                       value={values.occupants}
-                      placeholder="Total number of unit tenants"
+                      placeholder="Total number of unit tenants (Optional)"
                     />
                     {errors.occupants ? (
                       <div className="error-message">{errors.occupants}</div>
@@ -372,7 +373,7 @@ export const AddTenant = () => {
                         ? `${dateTimeStart.toDateString()} - ${dateTimeEnd.toDateString()}` 
                         : ""
                       }
-                      placeholder="Lease dates (Start and End)"
+                      placeholder="Lease dates (Optional)"
                     />
                     <CalendarModal title="Lease Range" calendarState={calendarState} iconYPosition="0.8rem" />
                   </div>
--- a/src/views/AddTenant/addTenant.js
+++ b/src/views/AddTenant/addTenant.js
@@ -23,7 +23,7 @@ const validationSchema = Yup.object().shape({
     .min(7, "*Phone Number must have at least 7 characters")
     .max(20, "*Phone Number can't be longer than 20 characters")
     .required("*Phone Number is required"),
-  unitNum: Yup.number(),
+  unitNum: Yup.string(),
   occupants: Yup.number(),
 });
 
@@ -39,11 +39,15 @@ export const AddTenant = () => {
   const [propertyOptions, setPropertyOptions] = useState([]);
   const [propertySearchResults, setPropertySearchResults] = useState([]);
   const [showAddProperty, setShowAddProperty] = useState(false);
+  const [isValidationActive, setIsValidationActive] = useState(false);
+  const [propertyErrorText, setPropertyErrorText] = useState("");
 
   const calendarState = useCalendarState();
-  const { dateTimeStart, dateTimeEnd, } = calendarState;
+  const { dateTimeStart, dateTimeEnd } = calendarState;
 
   useMountEffect(() => getProperties());
+
+  useEffect(() => validateForm(), [propertySelection, dateTimeStart, dateTimeEnd]);
 
   useEffect(() => {
     axios.post("/api/users/role", {
@@ -88,13 +92,8 @@ export const AddTenant = () => {
   };
 
   const handleFormSubmit = (data) => {
-    let body = {
-      ...data,
-      propertyID: propertySelection[0].key,
-      staffIDs: staffSelections && staffSelections.map(staff => staff.key)
-    };
     axios
-      .post(`/api/tenants`, body, makeAuthHeaders(context))
+      .post(`/api/tenants`, data, makeAuthHeaders(context))
       .then((response) => {
         Toast("Tenant Created Successfully!", "success");
       })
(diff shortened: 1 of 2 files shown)


aedwardg:
Looks good to me.

Although the unit number is not currently required on the backend, the other lease fields on this page (occupants and start/end dates) are.  I think until the backend provides functionality for _unhoused_ tenants, these changes will be super useful.  

Particularly with the upcoming demonstration, this will prevent accidentally omitting lease details and having the site hang (no start/end dates) or receiving a 400 error for "" being an invalid integer (no occupants) 😅

src/views/AddTenant/addTenant.js
```diff
@@ -23,8 +23,10 @@ const validationSchema = Yup.object().shape({
     .min(7, "*Phone Number must have at least 7 characters")
     .max(20, "*Phone Number can't be longer than 20 characters")
     .required("*Phone Number is required"),
-  unitNum: Yup.number(),
-  occupants: Yup.number(),
+  unitNum: Yup.string()
+    .required("*Unit Number is required"),
+  occupants: Yup.number()
+    .required("*Number of Occupants is required"),
```
NickSchimek: These shouldn't be required. Sometimes properties can be detached homes which would not have a unitNum. 
We're not requiring these fields in the backend. However, perhaps occupants needs to be required in the backend? We could check with @htharker42 on this.

src/views/AddTenant/addTenant.js
```diff
@@ -158,8 +178,13 @@ export const AddTenant = () => {
             occupants: "",
           }}
           validationSchema={validationSchema}
+          validate={() => {
+            setIsValidationActive(true);
+            validatePropertySelection();
+            validateLeaseDates();
```
NickSchimek: We probably only want to validate this if there is any data on these two forms; otherwise only validate the tenant form. Although, I'm not sure if that is what this is doing already?

src/views/AddTenant/addTenant.js
```diff
@@ -168,6 +193,12 @@ export const AddTenant = () => {
 
             setSubmitting(true);
             handleFormSubmit(toSubmit);
+            resetForm();
+            setPropertySearchText("");
+            setPropertySelection([])
+            setStaffSearchText("");
+            setStaffSelections([]);
+            setIsValidationActive(false);
```
NickSchimek: Wow this seems painful, I think I understand what it is doing. I find it weird that it is necessary. Perhaps one day I will learn React and have opinions about this :-)

NickSchimek (changes requested):
OK, take my review with a grain of salt. However, one thing I know for sure is that unitNum should not be required as it could be an ADU. See https://github.com/codeforpdx/dwellinglybackend/pull/158#discussion_r533019410

aedwardg:
@NickSchimek even though occupants isn't required on the backend, if you submit the form without a value for occupants, you get a 400 error that says "" is not a valid integer. This has a similar effect to requiring that field.

Is this maybe an issue with what the frontend is submitting for blank fields?

DaveTrost:
@NickSchimek and @aedwardg - I totally agree with what you both are saying. This PR is **not** changing the app to work the way it is meant to work. Instead, this code prevents errors during the demo.

I did have some interesting problems when submitting partial data. I didn't investigate the Python to the point of understanding (again, not the point of this PR) but I think sharing and documenting them here will help us make a better solution in the long run. So, here's what I tried (experiment results from Postman):

Success
- `occupants: null` - this works. Thank you @aedwardg for pointing this out.
- `unitNum: null` - this also works
- omit `dateTimeStart` and `dateTimeEnd` from the data - This works

Failure
- `dateTimeStart: null` - Yields a 400 error "Field may not be null." 
- `dateTimeEnd: null` - Yields a 400 error "Field may not be null." 

Interesting Dependencies
- `propertyID: null` and dateTimeStart/End omitted from the data (ie keys are not present in the JSON) - this works
- `propertyID: null` and dateTimeStart/End included in the data - this yields `"propertyID": ["Field may not be null."]`. AND a followup request yields a new error: `"message": "A tenant with this first and last name already exists"`
  - I think this means the lease creation failed, and the tenant creation continued. I would prefer to prevent committing the tenant entry to the db if bad lease data accompanies the tenant.

src/views/AddTenant/addTenant.js
DaveTrost: I'm going to try to modify the JSON data so that empty values are submitted as `null` instead of `""`. Then, I can remove the required setting for these properties.

DaveTrost:
Okay. I made some more changes. Still, this is far from perfect. The form needs custom validation - specifically to ensure that lease and property information are provided in conjunction with the other. It needs to be valid to submit a tenant without lease or property information. It also needs to be valid to include lease or property information with the tenant. For this case, both pieces of information are required. 
Note: the UI for the calendar modal does not have a "clear" option, so once the user starts to enter lease dates in the calendar, there's no going back to the starting point of the form to submit a tenant without a lease/property

NickSchimek:
> @NickSchimek even though occupants isn't required on the backend, if you submit the form without a value for occupants, you get a 400 error that says "" is not a valid integer. This has a similar effect to requiring that field.
> 
> Is this maybe an issue with what the frontend is submitting for blank fields?

Well, that is a valid validation. An empty string is certainly not an integer.

It is not uncommon for frontends to send an empty string when a field has not been filled in by the user. 

We can either have the frontend only send the data that was entered - this would fix the issue without having to make any changes in the backend. 

or

In the backend - we will need to setup the marshmallow schemas to not validate this field when it is an empty string and set it to None.

Either way is fine with me. 

Making the field required only patches the problem for someone else to discover and fix down the road.

NickSchimek:
Looks like there is a long standing discussion on empty strings in the Marshmallow repo:
https://github.com/marshmallow-code/marshmallow/pull/1381

NickSchimek:
@aedwardg Looks like @DaveTrost already fixed the issue by setting the values to `null` instead of empty string.  Nice work @DaveTrost

src/views/AddTenant/addTenant.js
DaveTrost: Ok. Now that I have a much better understanding of the backend .... the form will validate propertyID if lease dates are supplied. Lease dates are optional though, so propertyID is not required until after the lease dates are entered.

src/views/AddTenant/addTenant.js
```diff
@@ -158,16 +168,31 @@ export const AddTenant = () => {
...
           validateOnBlur={false}
-          onSubmit={(values, { setSubmitting }) => {
+          onSubmit={(values, { setSubmitting, resetForm }) => {
             const toSubmit = {
               ...values,
-              dateTimeStart,
-              dateTimeEnd,
+              occupants: values.occupants || null,
+              unitNum: values.unitNum || null,
+              propertyID: propertySelection[0].key,
```
aedwardg: When you try to submit a new tenant w/no lease through the UI, this line will cause the site to hang if a property is not selected.  The console error I received was a Formik TypeError, "propertySelection[0] is undefined"

I was able to get rid of this error and successfully create an "unhoused" tenant by changing the line to:
```suggestion
              propertyID: propertySelection[0] ? propertySelection[0].key : null,
```

I'll leave it to your discretion whether this is the best way to handle this, but we definitely should not have to select a property to create a tenant.

aedwardg (changes requested):
Great job getting this all fixed so quickly, @DaveTrost!

I only have the one suggested change.  Other than that, everything seems to work the way it should now as far as I can tell 😄

src/views/AddTenant/addTenant.js
NickSchimek: I agree. Property should be considered as part of the lease form. Tenants should not have a direct relationship with property, they're only assigned a property through a lease.

aedwardg (approved):

aedwardg:
@DaveTrost thanks for all the hard work on this! Your most recent commit does a good job handling the empty property field when no lease info is filled out. 🥳 

I did notice something else 👇 but I think it should probably be addressed in a different issue/PR so we can get this one merged.

---

If you select the lease dates, you get prompted to select a property - this is correct behavior.  However, the user is not blocked from hitting the Save button.  So, the post request goes through and a tenant gets created successfully, but the lease creation fails (as it should -- propertyID can't be null in a lease).

I don't know the best way to handle this from a UI/UX perspective, so I think maybe it should be discussed by people with more frontend knowledge.  Personally, I wouldn't be opposed to allowing this behavior (successful tenant creation, failed lease creation) so long as appropriate popups were shown. Users could always go in later and add a lease for the tenant.

For example:
- User could be prompted on on hitting Save that proceeding w/lease dates but no propertyID will result in tenant creation, but no lease.  
- Allow user to choose whether to go back or proceed anyway.
- If they proceed, show a green success popup for tenant and a red fail popup for lease.

Alternatively, you could just disable the save button once lease dates are filled out until either:
- a property is selected; or
- the dates are reset.

Again, I'm no expert in UI/UX, so this is probably better left to a separate frontend issue.

NickSchimek (approved):
Thanks @DaveTrost !! This is a big improvement.

PR #509, opened 2021-01-16
Rename property unit field

### LINKED ISSUE #508: Rename Property "unit" field in Frontend
**Describe the bug**
The frontend and backend differ when referencing the ill-named "Unit" field on Property. #487 is renaming this to "num_units" for clarity, so the UI should also respect that name change.
Backend issue is #487 

**To Reproduce**
1. Start both backend and frontend
2. Sign in as an admin
3. Go to Manage Properties
4. Click ADD NEW
5. Fill in the Property Information, including the "Units" field and click SAVE
6. Open database or use CURL GET request on the properties API endpoint to see no number of units recorded for the new property

**Expected behavior**
Property "num_units" field should be successfully saved to the backend and retrieved by the frontend

**Screenshots**

**Additional context**

Chris-Boe:
issue508

Closes #508 

Is a `npm install` necessary? n

### DIFF
--- a/src/views/Tenant/TenantView.js
+++ b/src/views/Tenant/TenantView.js
@@ -168,9 +168,9 @@ const Tenant = () => {
       comp: <div />,
     },
     {
-      key: "unit",
-      label: "Unit",
-      value: property.unit,
+      key: "num_units",
+      label: "Total Units",
+      value: property.num_units,
       inputType: "text",
       comp: <div />,
     },
--- a/src/views/Tenant/TenantView.js
+++ b/src/views/Tenant/TenantView.js
@@ -168,9 +168,9 @@ const Tenant = () => {
       comp: <div />,
     },
     {
-      key: "num_units",
+      key: "unitNum",
       label: "Total Units",
-      value: property.num_units,
+      value: tenant.lease && tenant.lease.unitNum,
       inputType: "text",
       comp: <div />,
     },
--- a/src/views/Tenant/TenantView.js
+++ b/src/views/Tenant/TenantView.js
@@ -169,7 +169,7 @@ const Tenant = () => {
     },
     {
       key: "unitNum",
-      label: "Total Units",
+      label: "Unit",
       value: tenant.lease && tenant.lease.unitNum,
       inputType: "text",
       comp: <div />,
--- a/src/views/addProperty/addProperty.js
+++ b/src/views/addProperty/addProperty.js
@@ -27,7 +27,7 @@ const validationSchema = Yup.object().shape({
   state: Yup.string()
     .max(50, `*State can't be longer than 50 characters`)
     .required(`*State is required`),
-  units: Yup.string()
+  num_units: Yup.string()
     .max(50, `*Unit can't be longer than 50 characters`),
   managers: Yup.array()
 });
@@ -112,7 +112,7 @@ export class AddProperty extends Component {
                     city: '',
                     state: '',
                     zipcode: '',
-                    units: ''
+                    num_units: ''
                   }}
                   validationSchema={validationSchema}
                   onSubmit={(values, { setSubmitting, resetForm }) => {
@@ -194,13 +194,13 @@ export class AddProperty extends Component {
                           <Field
                             className='column form-field'
                             type='text'
-                            name='units'
+                            name='num_units'
                             onChange={handleChange}
-                            value={values.units}
+                            value={values.num_units}
                             placeholder='Number of units'
-                            error={errors.units}
+                            error={errors.num_units}
                           />
-                          {errors.units ? (<div className='error-message'>{errors.units}</div>) : null}
+                          {errors.num_units ? (<div className='error-message'>{errors.num_units}</div>) : null}
                         </div>
 
                         <div className=' add-property__assign-manager-container'>


src/views/Tenant/TenantView.js
```diff
@@ -165,9 +165,9 @@ const Tenant = () => {
       comp: <div />,
     },
     {
-      key: "unit",
-      label: "Unit",
-      value: property.unit,
+      key: "num_units",
+      label: "Total Units",
+      value: property.num_units,
```
aedwardg: If I'm not mistaken, shouldn't this be the `unitNum` value of the lease?  Looking at page 10 of the [figma](https://drive.google.com/file/d/1YqboQogczYm1HkyRqEtVSzeQ61T9hWU2/view), I'd interpret this as the individual unit number connected to the tenant, rather than the number of units contained in the property.

aedwardg (commented):
Thanks for following through on the frontend as well for this name change @Chris-Boe!  I commented on `TenantView.js` because I think we want a different value there.  I'm not sure how we'd grab it though.  I suppose through tenant.leases and then find the current lease's `unitNum`.

src/views/Tenant/TenantView.js
Chris-Boe: Ah you're right! I'll update that. thanks for catching it!

src/views/Tenant/TenantView.js
```diff
@@ -165,9 +165,9 @@ const Tenant = () => {
       comp: <div />,
     },
     {
-      key: "unit",
-      label: "Unit",
-      value: property.unit,
+      key: "unitNum",
+      label: "Total Units",
```
aedwardg: I think "Unit" was probably the correct label for this one.
```suggestion
      label: "Unit",
```
This confusion is exactly why we didn't set this up right in the first place 😂

src/views/Tenant/TenantView.js
```diff
@@ -165,9 +165,9 @@ const Tenant = () => {
...
+      value: tenant.lease && tenant.lease.unitNum,
```
aedwardg, line 173: So currently, the Tenant model has a one to many relationship called `leases` rather than lease.  I was talking with @NickSchimek about it and we may change it to be a one to one called `lease`, in which case this would work.

**EDIT**: Based on Nick's comment below, this line will work and we'll just update how we send the lease data over from the backend 👍

aedwardg (changes requested):
@Chris-Boe sorry this issue is turning out to be a little more complicated than it seemed 😅 ,

I ran the app using both of your PRs and discovered a few things that were keeping this from working.

I included the frontend fixes in my suggested changes here. I also noted that we may want to discuss changing tenents to only have one lease, rather than multiple leases.

However, it looks like we'll need to change a couple things in the backend as well in order to pull the unitNum from the lease.

I got it working locally with the following backend fixes (which can be in a separate issue):
In `models/tenant.py` in the `json` method, add the following line:
```python
"leases": [lease.json() for lease in self.leases] if self.leases else [],
```

And in `models/lease.py` import `Time` and add in a `json` method to the `LeaseModel` class:
```python
from utils.time import Time
...
...
def json(self):
        return {
            "id": self.id,
            "propertyID": self.propertyID,
            "tenantID": self.tenantID,
            "occupants": self.occupants,
            "dateTimeStart": Time.format_date(self.dateTimeStart),
            "dateTimeEnd": Time.format_date(self.dateTimeEnd),
            "unitNum": self.unitNum,
        }
```

We could probably do this with Marshmallow serialization, but that was throwing a circular import error and I didn't have time to debug it tonight.

src/views/Tenant/TenantView.js
NickSchimek: Thanks for reviewing @aedwardg Here are some thoughts.

I think irregardless of how we have it setup in the backend. We should probably be only sending one lease to the frontend, unless they have a need for all leases. If the frontend uses `leases[0] then I don't think we could guarantee that the current or relevant lease will always be at index 0. So I would suggest that the frontend does not need to know how the relationship exists, but rather only needs to know that we are going to send them a json object with a lease nested inside a tenant. so the way @Chris-Boe has it should be fine.

For example. The backend will send this JSON object and will guarantee that the nested lease object contains one lease and is the relevant/current lease. So even though we may have a many-to-one relationship in the back we will be filtering out all the leases except for the relevant one and serialize that one inside the tenant object.
```
{ 
    tenant: {
        name: 'name'
        ...
        lease: {
            unitNum: 'd405'
            ...
        }
    }
}
```

aedwardg: Yep, good call @NickSchimek!  Looking at it again, I think you're right that the backend should handle this and just send the one lease up to the frontend.  I'll remove these suggested changes and open up a new backend PR for that.

Chris-Boe: Thanks @aedwardg and @NickSchimek !

aedwardg (approved):
Looks good @Chris-Boe, I think this is ready to merge. 🚀 

I must have missed the fact that @Protopaco already implemented most of my backend suggestions in [his PR](https://github.com/codeforpdx/dwellinglybackend/pull/195) before you rebased the [related backend PR](https://github.com/codeforpdx/dwellinglybackend/pull/200).

This should work now as-is so long as there is only one lease per tenant.  I'm going to open up an issue today for us to handle selecting the correct lease if a tenant has more than one.

Chris-Boe:
> Looks good @Chris-Boe, I think this is ready to merge. 🚀
> 
> I must have missed the fact that @Protopaco already implemented most of my backend suggestions in [his PR](https://github.com/codeforpdx/dwellinglybackend/pull/195) before you rebased the [related backend PR](https://github.com/codeforpdx/dwellinglybackend/pull/200).
> 
> This should work now as-is so long as there is only one lease per tenant. I'm going to open up an issue today for us to handle selecting the correct lease if a tenant has more than one.

Thanks! That does bring up an interesting point of lease history and whether or not we need to keep past tenant leases around for history/audit.
And yeah I saw a lot of the changes necessary were in that other PR so that helped get things working right!

PR #533, opened 2021-02-11
adds remove tenant functionality, by deleting the lease, and fetching…

Protopaco:
… property data

Allows user to delete a tenant's lease, thus removing them from the property.  This change only removes the lease, it does not remove the tenant.

It also removes the "getTenants()" function as that information now comes in with property object.

### DIFF
--- a/src/views/Property/PropertyView.jsx
+++ b/src/views/Property/PropertyView.jsx
@@ -47,6 +47,8 @@ const Property = () => {
   const [property, setProperty] = useState('');
   const [tenantArray, setTenants] = useState('');
   const [confirmChange, setConfirmChange] = useState(false);
+  const [confirmTenantRemoval, setConfirmTenantRemoval] = useState(false);
+  const [tenantToRemove, setTenantToRemove] = useState('')
   const [inputValues, setInputValues] = useState()
   const calendarState = useCalendarState(property?.dateTimeStart, property?.dateTimeEnd)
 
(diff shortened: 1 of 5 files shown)


src/views/Property/PropertyView.jsx
```diff
@@ -201,9 +189,20 @@ const Property = () => {
...
 
-    getProperty()
+  const handleTenantConfirmButton = () => {
+    const leaseId = tenantToRemove.lease.id;
+    axios
+      .delete(`${process.env.REACT_APP_PROXY}/api/lease/${leaseId}`)
+      .then(res => getProperty())
+      .then(setConfirmTenantRemoval(false))
+      .then(setEditingStatus(false))
+      .then(Toast("Tenant successfully removed"))
```
DaveTrost, line 204: Once the response is received (line 201), further calls are not dependent on the completion of those prior. In other words, this section needs only one ".then".

DaveTrost (changes requested):
After removing a tenant from the property, the tenant's detail page gives the following error: "Unhandled Rejection (TypeError): Cannot read property 'address' of undefined". I think a tenant without a property needs to be handled cleanly before merging this PR.

The remove button looks out of place in this app. Any ideas for giving it a look more consistent with the site design are welcome!

Something bothers me about the way edit mode works with the tenant removal feature. Probably fine for now, but I have ideas for the long run. I think removing tenants from the property is best done without needing to enter edit mode for the property. This will essentially separate property modifications (address, name, manager) from lease modifications (add/remove tenants).

src/views/Property/PropertyView.jsx
Protopaco: This has been corrected.  Thank you!

Protopaco:
> After removing a tenant from the property, the tenant's detail page gives the following error: "Unhandled Rejection (TypeError): Cannot read property 'address' of undefined". I think a tenant without a property needs to be handled cleanly before merging this PR.
> 
> The remove button looks out of place in this app. Any ideas for giving it a look more consistent with the site design are welcome!
> 
> Something bothers me about the way edit mode works with the tenant removal feature. Probably fine for now, but I have ideas for the long run. I think removing tenants from the property is best done without needing to enter edit mode for the property. This will essentially separate property modifications (address, name, manager) from lease modifications (add/remove tenants).

I agree about how TenantView handles Tenants without Leases.  It seems like it could with some conditional rendering.  Is that best handled through writing a new ticket?

I updated the remove tenant icon to make it more consistent with the app.

I agree about how the edit mode works.  It seems like editing the Property attributes, Property Managers, and Tenants could all be handled separately.  For this I tried to follow the FIGMA's layout, but I would be happy to break them out.  

Thanks!

DaveTrost:
> I updated the remove tenant icon to make it more consistent with the app.
> 

Do you know why the cursor effect on removing a property manager does not change to a pointer?

DaveTrost:
> I agree about how TenantView handles Tenants without Leases. It seems like it could with some conditional rendering. Is that best handled through writing a new ticket?

Yes it is! We talked about this in the development meeting and decided to open a separate ticket: https://github.com/codeforpdx/dwellingly-app/issues/534

src/views/Property/PropertyView.jsx
Chris-Boe: I'm not seeing this change

Chris-Boe (changes requested):
I think this just needs to be rebased and get merged conflicts resolved, as well as the change Dave requested done. I'm having trouble testing it again with all the updates both the frontend and backend have had

PR #540, opened 2021-02-25
mobile dashboard modules now tile vertically, not horizontally

### LINKED ISSUE #539: Mobile View - Admin dashboard
**Describe the bug**
The three widgets at the top top of the page want to stack horizontally in mobile view. 

**To Reproduce**
1. Go to http://localhost:3000/dashboard
2. Change browser to mobile view
3. Cringe

**Expected behavior**
Have them stack vertically so they are at least legible 

 - Device: all

**Additional context**
mobile view.

beekman:
Closes #539. Adds mobile specific admin CSS
None.
None.
- [X] Code builds successfully
- [X] Code has been tested and does not produce errors
- [X] Code is readable and formatted
- [X] There isn't any unnecessary commented-out code
- [X] There aren't any unnecessary commits or files changed (shouldn't touch the /stashed folder unless the issue requests it)

### DIFF
--- a/src/components/DashboardModule/dashboardmodule.scss
+++ b/src/components/DashboardModule/dashboardmodule.scss
@@ -1,5 +1,4 @@
 @import '../../scss/variables.scss';
-
 .dashboard {
     &__module {
         padding: 20px 20px 50px 20px;
@@ -64,5 +63,24 @@
     &__module_subtext {
         font-size: 14px;
         display: block;
+    }
+}
+
+@media screen and (max-width:600px){
+    .dashboard__modules_container {
+        display:block !important;
+        width: 100%;
+
+        .dashboard {
+            &__module {
+                padding: 20px 20px 50px 20px;
+                width: 100%;
+                position: relative;
+                display: block;
+                flex-basis: 100%;
+                flex-grow: 0;
+                color: $white;
+            }
+        }
     }
 }
--- a/src/components/DashboardModule/dashboardmodule.scss
+++ b/src/components/DashboardModule/dashboardmodule.scss
@@ -2,7 +2,6 @@
 .dashboard {
     &__module {
         padding: 20px 20px 50px 20px;
-        width: 30%;
         flex-basis: 30%;
         flex-grow: 0;
         color: $white;
@@ -66,7 +65,7 @@
     }
 }
 
-@media screen and (max-width:600px){
+@media screen and (max-width:920px){
     .dashboard__modules_container {
         display:block !important;
         width: 100%;
@@ -74,6 +73,7 @@
         .dashboard {
             &__module {
                 padding: 20px 20px 50px 20px;
+                margin-bottom: 20px;
                 width: 100%;
                 position: relative;
                 display: block;
--- a/src/scss/variables.scss
+++ b/src/scss/variables.scss
@@ -17,7 +17,7 @@ $white: #FFF;
 $black: black;
 $errorRed: red;
 
-$breakpointMobile: 1200px;
+$breakpointMobile: 600px;
 
 $gradient: linear-gradient(to bottom, #6f90fc 0%, #52baed 100%);
 
@@ -31,4 +31,3 @@ $info: $brandBlue;
 $dark: $gray;
 $button-disabled-background-color: $lightGray !important;
 $button-disabled-opacity: 1;
-
--- a/src/scss/variables.scss
+++ b/src/scss/variables.scss
@@ -17,7 +17,9 @@ $white: #FFF;
 $black: black;
 $errorRed: red;
 
-$breakpointMobile: 600px;
+$mobile: 600px;
+
+$breakpointMobile: 1200px;
 
 $gradient: linear-gradient(to bottom, #6f90fc 0%, #52baed 100%);
 


src/components/DashboardModule/dashboardmodule.scss
```diff
@@ -65,4 +64,23 @@
         font-size: 14px;
         display: block;
     }
-}
\ No newline at end of file
+}
+
+@media screen and (max-width:600px){
```
DaveTrost: Looks like a good spot for a variable of some sort - future mobile-responsive styling can build from the foundation you put in place here. 💪

src/components/DashboardModule/dashboardmodule.scss
```diff
@@ -65,4 +64,23 @@
...
+    .dashboard__modules_container {
+        display:block !important;
+        width: 100%;
```
DaveTrost, line 71: "Cardinal sin of CSS"

Is there an alternative way to accomplish this that avoids `!important`?

DaveTrost (changes requested):
Love the way the blocks stack up, it's a good look on small screens! 

I think that beginning here with global definitions for screen breakpoints is a good idea. I'm less concerned about the use of `!important` - especially if you can convince me the alternative solution is messy and/or convoluted.

src/components/DashboardModule/dashboardmodule.scss
beekman: > "Cardinal sin of CSS"
Well played, sir. I shouldn't have pull requested before cleanup. Took me a while to find the right hook and before that I was playing quick and loose with the overrides to save time.
I'm sure I can fix that.

src/components/DashboardModule/dashboardmodule.scss
beekman: So bulma has naming conventions for responsive flexbox column layouts that we aren't really taking advantage of in our DOM structure, and I think that might be cleaner still than defining our own mobile breakpoint. I think I'll try to incorporate bulma's columns in the DOM and take advantage of columns tiling vertically on mobile by default.

beekman: I've located an existing variable used by bulma at https://github.com/codeforpdx/dwellingly-app/blob/f17e909d8cd9eb2a7ace32daafeb02ed4d0f181b/src/scss/variables.scss#L20 that's used to define the mobile breakpoint. Currently, it's set to an unreasonably large number. In my mind, the mobile breakpoint should be set to something like 540px or maybe 720px, not 1200px as it is currently. I'll see if setting this number lower breaks or improves anything.

beekman: Unfortunately, there's a lot of custom SCSS that's overriding the thoughtful presets that bulma provides; for the most part we may as well not even be using bulma at all, given how inconsistently we're using its conventions. As it stands, the fastest way for me to patch all the weird custom css is with quick and dirty sass breakpoints, at least for now.

src/components/DashboardModule/dashboardmodule.scss
beekman: Unfortunately, there's a lot of custom SCSS that's overriding the thoughtful presets that bulma provides and making it difficult to know where exactly to find the offending styles that are screwing things up and which I would need to override with even greater css specificity. In other words, this is one of those situations where the ball was already all tangled up before I got here and the only way to avoid unraveling it is !important.

DaveTrost (dismissed):
Regarding the screen size breakpoint, I really do think we need to make 600px into a variable in variables.scss

It's a valid point about the existing mobile breakpoint. For now, I trust you to build a naming convention that can live alongside `breakpointMobile` and perhaps approximates some of the Bulma sensibilities.

@Chris-Boe - Do you have any ideas here?

Chris-Boe:
> Regarding the screen size breakpoint, I really do think we need to make 600px into a variable in variables.scss
> 
> It's a valid point about the existing mobile breakpoint. For now, I trust you to build a naming convention that can live alongside `breakpointMobile` and perhaps approximates some of the Bulma sensibilities.
> 
> @Chris-Boe - Do you have any ideas here?

It would be nice to have a breakpoint variable that we plan on using in other places!
Thanks @beekman for doing this, Bulma comes with some very strange CSS that has made all of us question why it was chosen as the CSS library for Dwellingly 😅 so we understand the struggle!

Chris-Boe:
The mobile dashboard looks great! I'm not seeing a left nav menu button though, but I don't think that was caused by these changes. I'll open an issue for it!

Chris-Boe (dismissed):

beekman:
> > Regarding the screen size breakpoint, I really do think we need to make 600px into a variable in variables.scss

I went ahead and modified the existing $breakpointMobile variable in bulma's variables.scss file to be set to 600px. (It was previously set to twice that, which makes no sense at all in terms of any mobile breakpoints I've ever dealt with.) Does this seem like a good solution? Or should I have set another variable for this purpose? @Chris-Boe, any thoughts?

Chris-Boe:
> > > Regarding the screen size breakpoint, I really do think we need to make 600px into a variable in variables.scss
> 
> I went ahead and modified the existing $breakpointMobile variable in bulma's variables.scss file to be set to 600px. (It was previously set to twice that, which makes no sense at all in terms of any mobile breakpoints I've ever dealt with.) Does this seem like a good solution? Or should I have set another variable for this purpose? @Chris-Boe, any thoughts?

That old breakpoint value definitely doesn't make any sense. As long as it isn't being used anywhere else I'm okay with overwriting it! But if anything else was relying on that old value, we'd want to keep it around and make a new breakpoint and maybe adjust the names of the variables.

beekman:
> That old breakpoint value definitely doesn't make any sense. As long as it isn't being used anywhere else I'm okay with overwriting it! But if anything else was relying on that old value, we'd want to keep it around and make a new breakpoint and maybe adjust the names of the variables.

We learned that this value was used elsewhere, so instead I set a new $mobile variable set to 600px. Thanks @Chris-Boe.

Chris-Boe (approved):

PR #541, opened 2021-02-27
Fix for issue 529

### LINKED ISSUE #529: Archived properties should not be able to be archived again
**Describe the bug**
Currently, on the Properties page, Archived and non-archived properties look the same and can be "archived again", even though nothing changes in the database.

**To Reproduce**
1. Go to Properties
2. Click on a property checkbox and click Archive Properties
3. Toggle the Show Archived to SHOW
4. Repeat the process to the same property

**Expected behavior**
* Disable the checkbox on already-archived properties so it can't be clicked
* Gray out the other column texts on already-archived properties but NOT the name field so it still looks clickable

**Screenshots**

**Additional context**

michelangelo17:
Closes #529 

Archived properties should not be able to be archived again 

Expected behavior from issue description: 

- [x] Disable the checkbox on already-archived properties so it can't be clicked
- [x] Gray out the other column texts on already-archived properties but NOT the name field so it still looks clickable

In order to make sure state updated correctly, made a change on how API request was handled and changed useMountEffect to normal useEffect which triggers anytime checkboxRenderCount updated. (Only happens in patch request for archival.) 

Could change the patch on the backend to no longer send the array of properties as a response since only a 200 ok is needed. 

Tests were manual, but no errors produced.

Feb-26-2021 20-57-02


Chris-Boe (changes requested):
Can we drop the change to the package-lock.json file since we are not updating any packages in this PR?

michelangelo17:
Oops! Sorry about that. Fixed!

Chris-Boe (approved):

PR #557, opened 2021-03-25
Tables are now responsive, vertically stacking table columns on mobile

beekman:
548, unusable tables on mobile.

None.

- [X ] Code builds successfully
- [ X] Code has been tested and does not produce errors
- [ X] Code is readable and formatted
- [X ] There isn't any unnecessary commented-out code
- [X ] There aren't any unnecessary commits or files changed (shouldn't touch the /stashed folder unless the issue requests it)


Chris-Boe (changes requested):
I think we'll still need to add some overflow-x css options to allow the tables to be horizontally scrolled:
image

beekman:
I think you should see something more like this, once I've updated the SCSS target to work for this structure as well as that used on the dashboard...
Screen Shot 2021-03-24 at 7 52 00 PM

htharker42:
image
It also looks like there may be some overflow issue we will want to address

beekman:
> It also looks like there may be some overflow issue we will want to address

Looks like I need to add a second target for the specific DOM structure used on that page, my SCSS target was probably too specific to catch that page. Thanks, I'll update the SCSS to vertically stack both table formats' rows.

beekman:
@Chris-Boe the Tickets module is filled with hard-wired widths on the table headers. Obviously, this approach doesn't work well with mobile. Could you see any problems with removing all of these from the views/tickets/tickets.jsx code?
```
headerStyle: () => {
      return { width: "20%" };
    }
```

Chris-Boe:
> @Chris-Boe the Tickets module is filled with hard-wired widths on the table headers. Obviously, this approach doesn't work well with mobile. Could you see any problems with removing all of these from the views/tickets/tickets.jsx code?
> 
> ```
> headerStyle: () => {
>       return { width: "20%" };
>     }
> ```

The percent widths should work for a responsive app.... since the percent applies to whatever is inherited, maybe there is something further up in the DOM tree that is setting the width and preventing the table from getting smaller?

Chris-Boe (changes requested):
We might want to take a different approach for the larger tables. The columns stacking on each other makes I think a  worse mobile experience than horizontal scrolling. It's especially confusing when some columns are missing data, so the next column's data fills left and looks mislabeled.
image

beekman:
> We might want to take a different approach for the larger tables. The columns stacking on each other makes I think a worse mobile experience than horizontal scrolling. It's especially confusing when some columns are missing data, so the next column's data fills left and looks mislabeled.

I have been thinking the same thing, but I'm glad I had you take a look first to confirm my impression that it was the wrong way to go. I'm going to return to my original idea to make the tables horizontally scroll like traditional responsive bootstrap tables for now. And down the road we can explore more fancy approaches that hide and show fields based on screen size and selection state.

PR #582, opened 2021-04-08
addNewAlias.md instructions updated with new command for re-seeding the database

### LINKED ISSUE #581: Instructions for creating a custom alias are incorrect
**Describe the bug**
Instructions for setting up an alias on addNewAlias.md are incorrect. They reference

**To Reproduce**
1. Follow the instructions on addNewAlias.md to create a new `dwellingly` alias for zsh.
2. Enter `source ~/.zshrc` in your Terminal on a mac, then return, then `dwellingly`.
3. The back end will attempt to run but choke on the outdated command for re-seeding the database.

**Expected behavior**
`dwellingly` alias should run the back end and then start the front end, all from one command.

 - OS: Macbook Pro running Big Sur

beekman:
- [X] Code builds successfully
- [X] Code has been tested and does not produce errors
- [X] Code is readable and formatted
- [X] There isn't any unnecessary commented-out code
- [X] There aren't any unnecessary commits or files changed (shouldn't touch the /stashed folder unless the issue requests it)

### DIFF
--- a/addNewAlias.md
+++ b/addNewAlias.md
@@ -1,8 +1,8 @@
-# Create a Custom Terminal Command 
+# Create a Custom Terminal Command
 
 ## Mac Instructions
 
-Once you get both the back-end and the front-end set up and working, you can do the following to create a terminal command `dwellingly` to re-seed the database, start the server, build the application, and launch it with all in one step by doing the follow. This is not required, but could save some time. This process can be used to set up other custom commands as well.
+Once you get both the back-end and the front-end set up and working, you can do the following to create a terminal command `dwellingly` to re-seed the database, start the server, build the application, and launch it with all in one step by doing the following. This is not required, but could save you some time. This process can be used to set up other custom commands as well.
 
 *Note that this has not been tested on Windows devices and the proccess will likely differ.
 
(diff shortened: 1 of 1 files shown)


NickSchimek:
Hmmm, I'm not sure what context this is being used in. It is not necessary to run db recreate before starting the server in the backend. Because that will wipe out any data you may have been working with from the last session. If db recreate should be left in then you'll want to install any new deps too. For ex: if the context this is being used in is after pulling down a fresh copy from the development branch. I would suggest either removing db recreate or adding pipenv run dev-install before db recreate.

beekman:
> Hmmm, I'm not sure what context this is being used in. It is not necessary to run db recreate before starting the server in the backend. Because that will wipe out any data you may have been working with from the last session. If db recreate should be left in then you'll want to install any new deps too. For ex: if the context this is being used in is after pulling down a fresh copy from the development branch. I would suggest either removing db recreate or adding pipenv run dev-install before db recreate.

I think in this case the latter makes more sense (adding `pipenv run dev-install` before db recreate), because the way I envision the command being used, it would be most helpful for those focused on the front end who might need to look up the commands to freshen everything up on the back-end server every time to make sure they're doing it right. Like, um, a friend of mine. So it should probably account for the more complicated sequence of commands that should follow a git pull, since those are the ones I... I mean, my friend... would tend to forget. However, the best solution might be two different aliases for the different scenarios.

Chris-Boe (approved):

Issue #579, opened 2021-04-08
Tenants should not be created with incomplete lease/property info


aedwardg:
**Describe the bug**
Right now you can add a tenant with: 
- incomplete lease info (filled in unit num and/or occupants, but missing dates), and no property info; or 
- property info but no lease info

The issue is that a tenant's property information is accessed through their lease to that property. 

Either of the above scenarios results in a successful tenant creation, but no lease creation (to make a lease though the `tenants` endpoint, a property ID and lease start and end dates must be included in the payload). 

Rather than having the lease creation fail silently, I propose we add a little more validation to the front end.
The current validation requires a property to be selected if the lease start and end dates don't match.  Instead, if _any_ lease/property info is filled out, the user should be notified that lease start/end dates and a property selection are required.

**To Reproduce**
1. Sign in as Admin
2. Go to 'Add New Tenant'
3. Add tenant info and select a property but include no lease info
4. Hit 'Save' and see successful tenant creation
5. Repeat, but this time, fill in the unit num and occupants, but no lease dates or property info

**Expected behavior**
User should be notified of required fields when property/lease info is added

**Additional context**
Right now, due to the problem described in #578, you cannot view tenant details for a tenant created this way because they won't have any lease/property info saved in the db

yunnie:
The add tenant view is found in `src/views/AddTenant/addTeant.jsx`

yunnie:
It looks like there is a form validation function `validateForm`

yunnie:
This is done and can be closed.

Issue #578, opened 2021-04-08
Cannot view unhoused tenant details


aedwardg:
**Describe the bug**
When trying to view an unhoused tenant's info in the TenantView, it results in a TypeError.

The issue is that the code to display property info doesn't check for the existence of the property assignment before trying to access it (`property` is `undefined` if the tenant has no lease):
https://github.com/codeforpdx/dwellingly-app/blob/5c11ab10a48ebad91bd792b1f67f5eeca0f88ae6/src/views/Tenant/TenantView.jsx#L192-L199

**To Reproduce**
1. Sign in to the app with an Admin account
2. Click on 'Add New Tenant'
3. Create a new tenant with no property or lease information
4. Go to 'Manage Tenants' and toggle to show Unhoused Tenants
5. Click on your new unhoused tenant.
6. See error.

**Expected behavior**
TenantView should allow user to see a tenant's info even if they have no lease/property.

**Screenshots**

**Additional context**
Other code that prevents this from working is the broken calendar code which tries to access `dateTimeStart` and `dateTimeEnd` through the property instead of the lease (see #580):
https://github.com/codeforpdx/dwellingly-app/blob/5c11ab10a48ebad91bd792b1f67f5eeca0f88ae6/src/views/Tenant/TenantView.jsx#L212-L213

yunnie:
The address, unit number and lease information will be empty for unhoused tenants. Will use "Not Housed" as a default value.

yunnie:
If there is no lease, then to display a "Not Applicable" in the lease section of the table, we need to make the `inputType` conditional, if there is a valid lease, use `calendar`, otherwise use `text`.

PR #584, opened 2021-04-11
fixed lease dates in manage tenant view

### LINKED ISSUE #580: Tenant info does not show correct lease dates
**Describe the bug**
In the TenantView, the lease start and end dates always show today's current date, regardless of the actual dates saved in the database.

**To Reproduce**
1. Sign in as Admin
2. Click on 'Manage Tenants'
3. Select a tenant that has a lease
4. See the improper dates

This is due to this code, which improperly tries to access `dateTimeStart` and `dateTimeEnd` through the tenant's property and ends up creating a new `Date` instance:
https://github.com/codeforpdx/dwellingly-app/blob/5c11ab10a48ebad91bd792b1f67f5eeca0f88ae6/src/views/Tenant/TenantView.jsx#L212-L213

**Expected behavior**
The lease start and end dates should reflect what is actually saved in the database.

**Screenshots**
image

yunnie:
Closes #580 and #578

- [ x] There isn't any unnecessary commented-out code
- [ x] There aren't any unnecessary commits or files changed (shouldn't touch the /stashed folder unless the issue requests it)

### DIFF
--- a/src/views/Tenant/TenantView.jsx
+++ b/src/views/Tenant/TenantView.jsx
@@ -209,8 +209,8 @@ const Tenant = () => {
       key: "lease",
       label: "Lease",
       value: {
-        dateTimeStart: property.dateTimeStart || new Date(),
-        dateTimeEnd: property.dateTimeEnd || new Date()
+        dateTimeStart: new Date(tenant.lease.dateTimeStart) || new Date(),
+        dateTimeEnd: new Date(tenant.lease.dateTimeEnd) || new Date()
       },
       inputType: "calendar",
       readOnly: true,
--- a/src/views/Tenant/TenantView.jsx
+++ b/src/views/Tenant/TenantView.jsx
@@ -192,7 +192,10 @@ const Tenant = () => {
     {
       key: "address",
       label: "Property",
-      value: `${property.address}, ${property.city}, ${property.state}, ${property.zipcode}`,
+      value: (property) ? 
+	`${property.address}, ${property.city},` + 
+	` ${property.state}, ${property.zipcode}` : 
+	"Unhoused Tenant",
       inputType: "text",
       comp: <div />,
       readOnly: true,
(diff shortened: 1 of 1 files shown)


aedwardg (dismissed):
Looks like this works!

yunnie:
This is related to #578. So I think it' better to just make one pull request that covers both issues.

devon-wolf:
I'm going through issues this morning, and I'm wondering:  does this also resolve #579 (tenants should not be created with incomplete lease/property info)?  

If this is assuming a default "not housed" status when a tenant is being created it seems like that issue becomes redundant/outdated. 

(If we'd like an explicit assignment  to 'unhoused' in the add tenant form, though, I'd be happy to update the issue and work on that. It would add clarity from the user perspective even if we're already handling it underneath.)

aedwardg:
@devon-wolf 

No, the issue in #579 is completely separate from this.  This PR (and its corresponding issue) are regarding the fact that lease dates displayed on the front end always showed today's date for both the lease start and end date -- even if the correct dates were stored on the back end.

The issue described in #579 has to do with lease creation and property assignment failing silently due to the front end allowing a user to create a new tenant without the information the back end needs for saving lease/property information to the database.

The front end _should_ (and does) allow a tenant to be created with either:
1. No property/lease information (unhoused tenant); or
2. With property/lease information (housed tenant)

The problem is that for option 2 (housed tenant), the following information _must_ be provided for the lease creation and property assignment to be successful:
- Lease start and end dates; **and**
- The property ID 

All other property/lease fields in the `TenantView` are optional.

Currently, when a user chooses to Add a Tenant, they can provide other lease/property options and neither of the above requirements, or just one of the above requirements and then hit "SAVE".  This results in a successful tenant creation, but the lease creation and property assignment never happen and the user has no idea that they were unsuccessful.

There are several ways to handle this; I suggested more form validation on the front end, but perhaps we should discuss it with Hugh at Wednesday's meeting to see what he prefers.

yunnie:
An alternative would be to just not return Property, Unit Number and Lease Dates at all. But I used "Unhoused Tenant" and "Not Applicable" (though it could also be "Not Available") to indicate that the information is missing.

Chris-Boe (approved):
Nice work!

Issue #530, opened 2021-02-11
Add ability to add notes to a ticket


Chris-Boe:
**Purpose**
What will this feature provide?
Ability to add notes to a ticket in the Ticket modal screen and see them show up in the list in real time.

**Proposed Solution**
If you have an idea on how to implement this feature, please share!
* Add a Plus sign button to the right of the Notes header
* Clicking the plus sign shows an input text box above the topmost existing note (and if they're scrolled down in the notes section, this will scroll them back up). Clicking the plus button also hides it from view until they add the note
* An "Add" button is displayed under the bottom right of the input box that saves the note to the backend. When this is clicked, the input box and Add button disappears and the Plus button shows up again.

**Acceptance Criteria**
Please list the requirements the implementation of this feature should meet.
- [ ] New Plus button displays input text box
- [ ] Clicking Save under the input box will save the new note to the backend

**Mockup Screenshot or Page Number**
Paste a screenshot of the feature from the FIGMA mockups or cite the page number.
![image](https://user-images.githubusercontent.com/8600792/107602493-9adcb780-6bde-11eb-8213-c26e398c6945.png)

**Additional context**

Protopaco:
I would be happy to make an attempt at this.
My plan is: 
1. Create an AddNote component that returns a text input and "add" button
2. add handleAddNote that will send PUT request to server with updated Note information
3. add "plus" button that changes a "showAddNote" state boolean
4. add turnery to display AddNote when plus button is pushed

Chris-Boe:
We'll probably need to make some updates to the backend too so lmk if you have any questions!

Protopaco:
> We'll probably need to make some updates to the backend too so lmk if you have any questions!

It looked like I could send a PUT request on the ticket with an updated array of notes with the current backend functionality.  I haven't tested it yet however.  Does that work or should I try a different approach?

Chris-Boe:
> > We'll probably need to make some updates to the backend too so lmk if you have any questions!
> 
> It looked like I could send a PUT request on the ticket with an updated array of notes with the current backend functionality. I haven't tested it yet however. Does that work or should I try a different approach?

Ah yep that looks like it'll work! It looks like we might need to update the "senderID" code though, since the person making the note might not always be the initial sender of the ticket. So with the request to add a note, we'll need to make sure there is an "authorID" or something that is only getting applied to the note model, not to the ticket model's SenderID

Protopaco:
[...]

PR #604, opened 2021-05-06
adds ability to add notes to tickets

Protopaco:
https://github.com/codeforpdx/dwellingly-app/issues/530

Is a `npm install` necessary?

### DIFF
--- a/src/components/AddNote/AddNote.scss
+++ b/src/components/AddNote/AddNote.scss
@@ -1,0 +1,28 @@
+@import "../../scss/variables.scss";
+@import "../card/Card.scss";
+
+.add-note-container{
+  display: flex;
+  flex-direction: column;
+  margin-bottom: 1em;
+
+}
+
+.add-note-input{
+  font-weight: normal;
+  padding: 1em .3em 1em 1em;
+  border-radius: $border-radius--card;
+  display: flex;
+  outline: none;
+  width: 100%;
+  color: $darkBlue;
+  background-color: $lightGray;
+  font-weight: normal;
+  font-size: 14px;
+}
+
+
+.add-note-add-button-div{
+  display: flex;
+  justify-content: flex-end;
+}
--- a/src/components/TicketModal/index.jsx
+++ b/src/components/TicketModal/index.jsx
@@ -40,7 +40,7 @@ export const TicketModal = (props) => {
             <Card.Content>
               <div className="card__summary">
                 <div className="close-icon-container">
-                  <button type="button" onClick={props.handleCloseTicket}>
+                  <button type="button" onClick={handleCloseTicket}>
                     <Icon id="close-icon" icon="close" />
                   </button>
                 </div>
--- a/src/components/TicketModal/index.jsx
+++ b/src/components/TicketModal/index.jsx
@@ -89,7 +89,7 @@ export const TicketModal = (props) => {
           <Card.Bottom>
             <Card.Content>
               <div className="ticket-card-bottom-header">
-                NOTES {notes ? notes.length : 0}
+                {` NOTES (${notes ? notes.length : 0})`}
                 <div onClick={toggleShowAddNote}>
                   <i className="fas fa-plus-circle icon-inline-space" />
                 </div>
--- a/src/components/TicketModal/index.jsx
+++ b/src/components/TicketModal/index.jsx
@@ -1,10 +1,12 @@
-import React, { Component } from "react";
+import React, { Component, useState } from "react";
 import Card from "../card/Card";
+import AddNote from "../AddNote/AddNote"
 import { CARD_TYPES } from "../../constants";
 import Icon from "../icon/Icon";
 import "./TicketModal.scss";
 
 export const TicketModal = (props) => {
+  const [showAddNote, setShowAddNote] = useState(false)
   if (!props.show || !props.ticket) {
     return null;
   }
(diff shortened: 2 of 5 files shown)


src/views/tickets/tickets.jsx
```diff
@@ -139,7 +140,7 @@ export class Tickets extends Component {
   }];
 
   getTickets = (context) => {
-    axios.get(`/api/tickets`, { headers: { "Authorization": `Bearer ${context.user.accessJwt}` } })
+    axios.get(`/api/tickets`, axios.get(`/api/tickets`, makeAuthHeaders(context)))
```
Chris-Boe: Looks like an extra axios call got in here somehow and is breaking the page load occasionally

src/views/tickets/tickets.jsx
```diff
@@ -285,16 +300,17 @@ export class Tickets extends Component {
...
                       })}
                     />
                   </div>
                 </div>
                 <TicketModal
                   show={this.state.viewedTicket}
                   onClose={this.toggleTicketModal}
-                  ticket={this.state.viewedTicket}>
-                </TicketModal>
+                  ticket={this.state.selectedTicket}
```
Chris-Boe: The modal is not showing up because there is no "selectedTicket" property in state. This should be "this.state.viewedTicket" like before

src/views/tickets/tickets.jsx
```diff
@@ -160,6 +161,20 @@ export class Tickets extends Component {
     });
   };
 
+  handleAddNote = (noteText, ticketID) => {
+    axios.post(`/api/tickets/${ticketID}/notes`, { text: noteText }, makeAuthHeaders(this.context))
+      .then(({ data }) => {
+        const updatedSelectedTicket = this.state.selectedTicket;
+        updatedSelectedTicket.notes.push(data)
+        this.setState({ selectedTicket: updatedSelectedTicket })
```
Chris-Boe: "selectedTicket" should be "viewedTicket" I believe

src/views/tickets/tickets.jsx
```diff
@@ -195,13 +210,13 @@ export class Tickets extends Component {
       },
       headers: { "Authorization": `Bearer ${this.context.user.accessJwt}` }
```
Chris-Boe: Let's use that makeAuthHeaders method here too

src/components/TicketModal/index.jsx
```diff
@@ -28,7 +40,7 @@ export const TicketModal = (props) => {
             <Card.Content>
               <div className="card__summary">
                 <div className="close-icon-container">
-                  <button type="button" onClick={props.onClose}>
+                  <button type="button" onClick={props.handleCloseTicket}>
```
Chris-Boe: "handleCloseTicket" is not a **prop** given to this modal, just a method defined above, so this should just be 
```onClick={handleCloseTicket}```

**Review by Chris-Boe — CHANGES_REQUESTED** [review]

File: src/components/AddNote/AddNote.scss [hunk 6]
```diff
@@ -0,0 +1,28 @@
+@import "../../scss/variables.scss";
+@import "../card/Card.scss";
+
+.add-note-container{
```
Chris-Boe, line 4: We'll also want to be able to have a scrollbar for the notes container, so users can scroll down to view all of them

src/components/TicketModal/index.jsx
```diff
@@ -78,7 +90,17 @@ export const TicketModal = (props) => {
...
               </div>
+              {showAddNote ?
+                <AddNote
+                  handleAddNote={props.handleAddNote}
+                  toggleShowAddNote={toggleShowAddNote}
+                  ticketID={id}
+                />
+                : null}
               {notes ? (
                 notes.map((note) => {
```
Chris-Boe: Do we want to order these with the most recent on the top? It might make more sense since the input field is above the list

Chris-Boe (changes requested):
Looks great! Thanks for making those other changes, just had a couple thoughts on the notes section's UI

src/components/TicketModal/index.jsx
Protopaco: We can order the notes with the most recent on the top, but I feel like that would probably be a backend thing.  I assume we can change the order in which they are take from the database, rather than reorder them in the frontend, if that makes sense.  I can create a ticket for that if you'd like.  Thanks!

src/components/AddNote/AddNote.scss
Protopaco: Just want to clarify.  Do you mean to add a scroll bar so the existing notes container so the user can view all of the notes while adding a note?  I think that would probably be a good thing to add, since currently the ticket modal doesn't appear to have any handling for that.  

I played around with some changes to the Ticket Modal and Ticket Modal.scss that I think achieves this goal.  Sorry about all of the little commits.

src/components/TicketModal/index.jsx
Chris-Boe: Yeah that would be great! Thanks

src/components/AddNote/AddNote.scss
Chris-Boe: It looks great when there are many notes on a Ticket, but when there is only a couple notes, we might want to adjust the spacing and remove the scrollbar until it is needed.

In the #scroll-container style rule, let's change overflow-y to 'auto'.

image

Chris-Boe: Also do you mind adding parentheses around the number of notes? I think it's clearer to have Notes (5) vs Notes 5

Chris-Boe (approved):
Thanks for the great work!

PR #643, opened 2021-05-27
Form validates dates and property

yunnie:
#579 

- [x ] Code builds successfully
- [x ] Code has been tested and does not produce errors
- [x ] Code is readable and formatted
- [x ] There isn't any unnecessary commented-out code
- [x ] There aren't any unnecessary commits or files changed (shouldn't touch the /stashed folder unless the issue requests it)

### DIFF
--- a/src/views/AddTenant/addTenant.jsx
+++ b/src/views/AddTenant/addTenant.jsx
@@ -150,7 +150,7 @@ export const AddTenant = () => {
     }
     if (dateTimeStart !== dateTimeEnd) {
       if (!propertySelection.length) {
-        errors.propertySelection = "Property is required when lease dates are slected"
+        errors.propertySelection = "Property is required when lease dates are selected"
       }
     }
     if (values.unitNum || values.occupants) {
--- a/src/views/AddTenant/addTenant.jsx
+++ b/src/views/AddTenant/addTenant.jsx
@@ -1,6 +1,6 @@
 import React, { useContext, useEffect, useState } from "react";
 import { Link } from "react-router-dom";
-import { Form, Field, Formik } from "formik";
+import { Form, Field, Formik, ErrorMessage } from "formik";
 import * as Yup from "yup";
 import * as axios from "axios";
 import UserContext from "../../UserContext";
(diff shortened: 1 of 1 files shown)


yunnie:
Changed the validation so that all validation is done via Formik. If a property is selected the lease start and end dates will be required. If lease start and end dates are provided, then a property is required.

Removed some state hooks that are no longer necessary because all validation is done via Formik.

Also used the Formik `ErrorMessage` component to display any validation errors.

aedwardg:
@yunnie, 

I really like what you did with this and I think it's a big improvement already.
However, it looks like it still doesn't cover the first problem I mentioned in the issue:
> incomplete lease info (filled in unit num and/or occupants, but missing dates), and no property info

If the user puts in a unit number/occupants but leaves the lease dates and property blank, the form still submits, creating a tenant and failing silently on the lease. 

I guess the question is: is this what we want to happen? Or do we want to prevent the form from submitting? Or do we want to warn the user that the lease won't be created if they continue?

yunnie:
It does make sense to assume that if the user enters a Unit Number of a Number of Occupants that they intend to enter a lease, in which case a property and lease start and end dates are also required. If the entry of unit number and/or number of occupants was accidental, then validation can alert the user to the accidental entries (via error messages stating that property and lease dates are also required). I'll make the change to the validation.

src/views/AddTenant/addTenant.jsx
```diff
@@ -143,16 +138,35 @@ export const AddTenant = () => {
...
+        errors.lease = "Lease dates required when a property is selected"
+      }
+    }
     if (dateTimeStart !== dateTimeEnd) {
-      setPropertyErrorText(dateTimeStart !== dateTimeEnd && propertySelection.length
-        ? "" 
-        : "*Property is a required field"
-      );
+      if (!propertySelection.length) {
+        errors.propertySelection = "Property is required when lease dates are slected"
```
aedwardg:
```suggestion
        errors.propertySelection = "Property is required when lease dates are selected"
```

aedwardg (commented):
@yunnie This is great. It's a very graceful way to handle the validation and I feel pretty confident that after this change even the most clueless of users will be able to properly fill out this form.

I did find one tiny typo, but other than that, this looks good to go

aedwardg (approved):
Looks good to me!

Chris-Boe (approved):
Nice work!
I noticed a couple things were broken that were not part of this PR so I'll create separate issues for them

PR #655, opened 2021-06-21
Set up system tests

NickSchimek:
## Description:
- Setup end to end system tests that run on the frontend and backend. 

### What this does:

**For CI Runs:**

1. Pulls down the latest backend development code (development branch) and spins up the server
2. Checks out the current branch for the frontend (PR branch) and spins up the react server
3. Runs all system tests using [Cypress](https://www.cypress.io/)

**For Local Runs**
Same as CI, but you need to start the servers yourself before executing the run command. 
To run Cypress locally. This should work `npx cypress run`
That should run in a headed mode. If it doesn't you can force headed mode with `npx cypress run --headed`
To run in headless mode `npx cypress run --headless`
see: https://docs.cypress.io/guides/guides/command-line#cypress-run

The API docs are here: https://docs.cypress.io/api/table-of-contents

#### This does not: 
This does not run the unit tests that have been created in the frontend. 
These are still needed and can be ran as normal.
Furthermore, I believe Ryan is setting up the CI to run those.

#### If this looks OK: 
- **I will setup the backend to run all PR's with the frontend.**

We'll have to write system tests, but once written you wont have to manually test anymore. And this should prevent any future over sights, and make life easier........ 

The goal should still be to continue writing unit tests, and on top of that write systems tests to check user interaction.

Let me know what you think @htharker42 and @Chris-Boe and @others

I've written one test. User Login - Here it is in action

cypress_test_run

This is what the test looks like in code:
```js
describe('Login', function () {
  it('logs in the user', function () {
    cy.visit('http://localhost:3000/')
    
    cy.get('input[name="email"]').type('user1@dwellingly.org')
    cy.get('input[name="password"]').type('1234')
    cy.get('button').contains('LOG IN').click()
    
    cy.location('pathname').should('eq', '/dashboard')
  })
})
```


Protopaco:
This is very cool, I hadn't heard of this before.  Would this work with or replace the backend integration tests?  I'm up for giving it a try, so long as it doesn't slow down the rest of development.  I feel like we are getting close!

NickSchimek:
> Would this work with or replace the backend integration tests?

This works with all existing tests, and does not replace any existing tests. The integration tests in the backend are almost unit tests. They're kinda placed in a badly named folder. They only test parts of the backend. 

> I'm up for giving it a try, so long as it doesn't slow down the rest of development.

Yes, as long as it doesn't slow down development. That depends on how the tests are written. These are system tests. We don't need a lot of system tests. Just a few to verify that any changes made in a PR is not going to break the website.

We need/should have system tests, but these are not to replace any existing tests, and should not slow down development as long as care is taken when writing them.

System tests take longer to run, but we only need a few here and there to verify that future PR's and upgrades will not bring down the site. It should also help alleviate a lot of the manual testing. 
For instance: when we updated the jwt package in the backend, we knew it was a breaking change and had missed one spot in the frontend that needed to be changed. This one test would've caught that. This one test runs a lot of code, hence the reason why we don't need a lot of system tests. Once you start over writing system tests then it starts to bog down development, because every change is going to cause a lot of system test failures. So care must be taken. And there are quite a few of patterns that we can take advantage of in Cypress to make it easy to update the tests when there are changes that we know will cause failures.

aedwardg (dismissed):
This looks good to me! I think it will be very useful for us to have a few end-to-end tests.
Between this and the staging/QA deployment we discussed setting up last meeting I think we'll be able to catch any remaining or newly added bugs much faster.

NickSchimek:
I must’ve missed that discussion

NickSchimek:
> This looks good to me! I think it will be very useful for us to have a few end-to-end tests.
Between this and the staging/QA deployment we discussed setting up last meeting I think we'll be able to catch any remaining or newly added bugs much faster.

Thanks. I'd be happy with just this one test and not many more. This application is quite small and there is not much to test in regards to system tests. This one test checks a lot of the main things that could come up, and a couple more that checks essential parts of the app would be good too. As long as we have the frontend/backend covered by unit/integration tests then that will allow us to keep system testing to a minimal. 

As far as a staging and QA deployment. I'd be onboard for a staging environment, but I don't think it's really needed right now. of course it doesn't hurt to have one though, but I'm not really seeing a benefit to having one right now. I used to think it was always a good idea, but I've kinda changed my tune.

I've honestly set this up because it is the second or third time I got a PR in the backend that required me to verify the system, so I am automating that process so that I can spend more time doing other things, because while system tests are slow, manual testing is even slower (and I have to be present for manual testing).

.github/workflows/system_test.yml
```diff
@@ -0,0 +1,62 @@
...
+        working-directory: backend
+
+      - name: Setup Database
+        run: pipenv run flask db create
+        working-directory: backend
+
+      - name: Start Server
+        run: pipenv run flask run &
+        working-directory: backend
+      - name: Just wait a sec
```
Chris-Boe, line 56: 😁

Chris-Boe (dismissed):
This is incredibly helpful to have it setup! Thanks @NickSchimek ! We can start creating issues to have additional tests setup - I'd say our base goal would be to verify each page opens without issue. We could test more workflows, like CRUD operations on several pages, but that large set of test criteria could be saved for Staging validation before prod deployments

NickSchimek:
Thanks Chris, yeah I agree a good set of tests would be to test everything the user clicks, which would cover the cases of all the pages opening and crud routes. And then save more fine grained testing for unit tests. We can always re-evaluate and adjust our testing strategy as we move along.

I will get the postgres pr merged in the backend and set this up to run against the postgres db too.

Chris-Boe:
> Thanks Chris, yeah I agree a good set of tests would be to test everything the user clicks, which would cover the cases of all the pages opening and crud routes. And then save more fine grained testing for unit tests. We can always re-evaluate and adjust our testing strategy as we move along.
> 
> I will get the postgres pr merged in the backend and set this up to run against the postgres db too.

Sounds good, @NickSchimek . I think we're ready to merge this whenever you are

scybert:
y not have 1 repo?

scybert:
a monorepo might help. especially with system tests.

.github/workflows/system_test.yml
scybert: pirate talk. avoid.

NickSchimek:
@scybert I'm not sure what factors were taken into consideration when the repo's were created. I've done projects where they're split into two repo's and others where the frontend and backend are all in the same repo. I think it mainly impacts deployments, and possibly developer experience.

I personally prefer to keep all application logic together in a single repo, but then again that really depends on the project itself. As far as this project is concerned 🤷 , I don't know. I wasn't around at the beginning for those decisions, so I am not the right person to answer your questions. I would also prefer to not reach for React as a frontend tool, unless I could achieve something with it that I couldn't easily do with other tools. Lots of ways to solve a problem, and most likely never one right way to solve it.

As far as system tests go, I don't think it matters if the backend is in the same repo or not. We still need system tests, and they would still behave the same way.

How do you see having the code in a single repo being beneficial for system tests?

PR #661, opened 2021-07-11
Issue658 Fix role API calls

### LINKED ISSUE #658: Remove backend call to get list of roles
**Purpose**
Remove the backend call to request a list of roles. This information is readily available in the frontend code and an ajax request is not needed here.

The frontend has a RoleEnum that can be used to get a list of roles. The following ajax requests should be replaced with a method call to the FE RoleEnum.

The following needs to be updated:

https://github.com/codeforpdx/dwellingly-app/blob/432888bcac2ef5362719ed668bd11f00732ef63d/src/views/requestAccess/requestAccess.jsx#L71

Screen Shot 2021-07-07 at 9 30 44 AM

Chris-Boe:
Closes #658 

Is a `npm install` necessary? n

### DIFF
--- a/src/views/Manager/index.jsx
+++ b/src/views/Manager/index.jsx
@@ -88,7 +88,7 @@ const Manager = () => {
 
   const updateManager = (payload) => {
     axios
-      .patch(`${process.env.REACT_APP_PROXY}/api/user/${id}`,
+      .patch(`/api/user/${id}`,
         payload, makeAuthHeaders(userContext)
       )
       .then(response => {
--- a/src/views/Manager/index.jsx
+++ b/src/views/Manager/index.jsx
@@ -145,7 +145,7 @@ const Manager = () => {
             <div className="section-container">
               <h2 className="secondary-title">PROPERTIES</h2>
               <div className="manager__properties__container">
-                {managerData.properties.map((property) => (
+                {managerData.properties?.map((property) => (
                   <div key={property.id} className="manager__property__tile">
                     <h3 className="manager__property__name">
                       {property.name}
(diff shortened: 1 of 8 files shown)


src/views/requestAccess/requestAccess.jsx
```diff
@@ -68,20 +68,11 @@ export const RequestAccess = (props) => {
...
-      })
-      .catch((error) => {
-        Toast(error.message, "error");
-        console.log(error);
-      });
+    let roles = [];
+    for(var key in RoleEnum) {
+      roles.push(key);
+    }
+    setSelectionOptions(roles);
```
NickSchimek: Would this work too?

```suggestion
    setSelectionOptions(Object.keys(RoleEnum));
```

src/views/Manager/index.jsx
```diff
@@ -145,7 +145,7 @@ const Manager = () => {
             <div className="section-container">
               <h2 className="secondary-title">PROPERTIES</h2>
               <div className="manager__properties__container">
-                {managerData.properties.map((property) => (
+                {managerData.properties?.map((property) => (
```
NickSchimek, line 148: What does this `?` mark do?

NickSchimek (dismissed):
Thanks for doing this Chris! I have a couple questions, otherwise it looks good! Thanks!

src/views/Manager/index.jsx
Chris-Boe: It's an optional chain marker, meaning if managerData.properties is "nullish", it will short circuit and will just return undefined instead of throwing an error. Since this is in the render method, it falls back to rendering nothing

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/views/requestAccess/requestAccess.jsx
Chris-Boe: Much better! I think I was way too tired when I was writing this fix 😂

src/views/Manager/index.jsx
NickSchimek: Nice! I didn't know Javascript had that. Ruby has something similar called safe navigation using the `&` character.

NickSchimek (approved):

PR #684, opened 2021-08-30
Remove race condition in updateProperty async functions

### LINKED ISSUE #677: Fix state management in the Properties component
**Describe the bug**
When a property is updated Then page does not get re-rendered to show the updated fields. It looks like the component doesn't always re-render when the state changes. 

**To Reproduce**
This doesn't occur all the time, but usually occurs on the first update after visiting the properties page and attempting to update a field on the form. After updating you need to hit refresh in order to see the change.

**Expected behavior**
The page should show the updated fields after submitting an update

NickSchimek:
The `handleConfirmButton` awaits on the `updateProperty` function to
update the property, and then does an additional call to get the
property. The issue is that the await keyword doesn't really impact
anything here, because the `updateProperty` function is also an async
function. So what was happening was that sometimes the property would be
updated with stale information due to the `updateProperty` and
`getProperty` racing to fetch data and update the state.

Removed the `getProperty` call as it is not needed. The `updateProperty`
method already updates the state.

Fixes #677

### DIFF
--- a/src/views/Property/PropertyView.jsx
+++ b/src/views/Property/PropertyView.jsx
@@ -100,7 +100,6 @@ const Property = () => {
   const handleConfirmButton = async () => {
 
     await updateProperty(inputValues);
-    getProperty();
     setConfirmChange(false);
   }
 


Chris-Boe (approved):
Simple enough!

PR #715, opened 2021-10-17
Add testing section to Contributing doc

NickSchimek:
Updated the contributing doc to include a section about how to run tests

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -122,7 +122,7 @@ Jest is used for unit testing. These can be ran with `npm run test`
 To run the system tests both the backendend server and frontend server must be running
 The backend server must be seeded with the `minimal_seed` data file.
 
-To ensure your db is seed correctly run the following commands:
+To ensure your db is seeded correctly run the following commands:
 1. `pipenv run flask db drop`
 2. `pipenv run flask db create`
 3. `pipenv run flask db minimal_seed`
--- a/contributing.md
+++ b/contributing.md
@@ -9,6 +9,7 @@ This document provides some overall guidelines and suggestions for how to get st
 - [Build a Feature Branch](#feature-branch)
 - [Make a Pull Request](#pull-requests)
 - [Where to Contribute](#where-to-contribute)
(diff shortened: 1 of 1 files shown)


contributing.md
```diff
@@ -104,3 +107,34 @@ The app is not yet deployed and this opens a whole realm of additional contribut
...
+Jest is used for unit testing. These can be ran with `npm run test`
+
+#### System Tests
+
+[Cypress](https://www.cypress.io/) is used for system testing.
+
+To run the system tests both the backendend server and frontend server must be running
+The backend server must be seeded with the `minimal_seed` data file.
+
+To ensure your db is seed correctly run the following commands:
```
Chris-Boe:
```suggestion
To ensure your db is seeded correctly run the following commands:
```

contributing.md
```diff
@@ -104,3 +107,34 @@ The app is not yet deployed and this opens a whole realm of additional contribut
...
+1. `pipenv run flask db drop`
+2. `pipenv run flask db create`
+3. `pipenv run flask db minimal_seed`
```
Chris-Boe, line 128: This can all just be `pipenv run flask db recreate` right?

Chris-Boe (dismissed):

contributing.md
NickSchimek: No, recreate will seed the db using the large random seed generation file, which will cause the tests not to pass, because there will be unexpected data returned.

The minimal_seed only creates an admin user.

Cypress has some way to create data in the db I believe, but I haven't looked into it yet. This may change in the future, but for now this is working.

NickSchimek: It would be nice to run the system tests against the test_database so that the dev database can be left alone, but again maybe that can be done in the future.

Chris-Boe: Ah makes sense! Thanks for the explanation. And yeah we can probably get them setup to run on the test environment instead at some point

Chris-Boe (approved):

PR #717, opened 2021-10-17
Fix inviting a user (add property manager/staff)

### LINKED ISSUE #678: An invited user is not granted authorization to the app
**Description**

The UserInvite resource in the backend will always create a User of type User. Meaning that they are not an authorized user until approved. The frontend uses this to create a property manager, and possibly other uses as well.

When used to create a property manager the FE provides the role. And since this is an Admin only function it makes sense that the admin would not need to approve the user after inviting the user.

Currently in the Invite we always create the User with `UserModel`, however if we are given a role in the payload then we should create that type of user as provided by the role. 

One option would be to create the user using the role that is passed. For instance if given a property-manager role then we should create the user with the PropertyManager class. 

Another option would be to create the user using the UserModel class, but assign the type of the user when creating the user.

NickSchimek:
This passes the type field along with the request which is necessary for
the user/invite endpoint. Eventually will be able to remove the role
field, but for now it is still required.

Fixes #678 

Not necessary, but a re-seeding of the db will ensure that there is a good data in the database for manually testing since this fixes an issue with bad data being populated to the db.

### DIFF
--- a/.github/workflows/system_test.yml
+++ b/.github/workflows/system_test.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Seed db
         run: |
           pipenv run flask db create
-          pipenv run flask db seed
+          pipenv run flask db minimal_seed
 
       - name: Start Server
         run: pipenv run flask run &
--- a/cypress.json
+++ b/cypress.json
@@ -1,1 +1,3 @@
-{}
+{
+  "baseUrl": "http://localhost:3000/"
+}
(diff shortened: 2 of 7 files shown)


Chris-Boe (approved):
Tested with PR 358 in the backend and everything seems good! Thanks for writing the test

NickSchimek:
Cool Thanks! This should work with the dev branch in the backend. It's not related to any backend pr.
And this kinda sets the stage for switching to User types instead of roles.

PR #728, opened 2021-10-29
Format dates to user's time zone Issue #605

### LINKED ISSUE #605: Show all times in the user's timezone
**Purpose**
What will this feature provide?
All times throughout the app will be shown in the user's timezone so it is easier to read

**Proposed Solution**
There are times displayed on the Manage Properties, Manage Property Managers, and Tickets pages that will need to be displayed in the user's timezone instead of UTC. MomentJS is a popular JS library for this as a starting point, but as of late 2020, it is no longer being worked on.

**Acceptance Criteria**
- [ ] Dates are shown in the user's timezone instead of UTC

Paste a screenshot of the feature from the FIGMA mockups or cite the page number.

**Additional context**

RyRojas:
Closes #605 

Is a `npm install` necessary? N
Any special requirements to test? N

Commits might be a bit much for something so simple. Let me know if you have any concerns with the changes made.

### DIFF
--- a/src/utils/formatDate.js
+++ b/src/utils/formatDate.js
@@ -1,0 +1,2 @@
+const formatDate = dateString => new Date(dateString + ' UTC').toLocaleString();
+export default formatDate;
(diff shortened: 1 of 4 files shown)


NickSchimek (approved):
I think this is good. This will also allow the backend to remove the special formatting we're applying on these date timestamps and it will work seamlessly.

toLocaleString without any arguments passed to it uses a format in the default locale with default options. I'm assuming the default locale would be set by the clients computer.

And in the future if we want the user to be able to set it explicitly then we will still be able to do that with toLocaleString.

Thanks! I think these changes look good

PR #742, opened 2021-11-13
Make catching the rejected API call optional

### LINKED ISSUE #741: Attempting to create a property with a name that already exists results in an unhandled promise error
**Describe the bug**
Attempting to create a property with a name that already exists results in an unhandled promise error

**To Reproduce**
1. Create a property with a name that already exists

**Expected behavior**
No errors

**Screenshots**
Screen Shot 2021-11-14 at 8 35 39 AM

NickSchimek:
Fixes #741

Any special requirements to test? dont think so

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -201,6 +201,11 @@ export class App extends React.Component {
       Toast(toastMessages.error ?? error.message, "error");
       return Promise.reject(error);
     })
+    // since we already display the erorr message. The caller does not need to catch the returned Promise.
+    // catching the rejected Promise is optional. so catch for them and return false.
+    .catch(error => {
+      return false
+    })
   );
 
   render() {


src/App.jsx
```diff
@@ -201,6 +201,11 @@ export class App extends React.Component {
       Toast(toastMessages.error ?? error.message, "error");
       return Promise.reject(error);
     })
+    // since we already display the erorr message. The caller does not need to catch the returned Promise.
```
Chris-Boe, line 204: Should we just replace the `return Promise.reject(error)` above with `return false`?

I thought about doing something like this but wanted the error to be available to the caller for additional catching, but I think this is fine

Chris-Boe (approved):

src/App.jsx
NickSchimek: > Should we just replace the return Promise.reject(error) above with return false?

I thought about this too, and almost did that exact thing.

But then I came to the same conclusion as you. I think it will be useful to have the error available to the caller.

PR #750, opened 2021-11-17
protect destructure from null error

### LINKED ISSUE #748: TypeError in Admin panel due to unsafe destructure
**Describe the bug**

**To Reproduce**
1. Clear a new user in admin panel
2. Page refreshes as blank, attempts to destructure items from a null `props.location.state`

**Expected behavior**
Site does not crash :)

**Screenshots**
Screen Shot 2021-11-03 at 8 06 48 PM

 - OS: OSX
 - Browser Chrome
 - Version 95.0.4638.69

zacksabbath:
Error I encountered in admin panel that led page to crash

Any special requirements to test? Admin panel should not crash when clearing a new user

### DIFF
--- a/src/views/RequestAccess/index.jsx
+++ b/src/views/RequestAccess/index.jsx
@@ -49,7 +49,7 @@ export const RequestAccess = (props) => {
     lastName,
     email,
     id
-  } = props.location.state;
+  } = props.location.state || {};
 
   const [selectionOptions, setSelectionOptions] = useState([]);
   const [roleSelection, selectionHandler] = useState("");


NickSchimek:
I don't understand why this change is necessary. What kind of error did you see? And how can I reproduce it?

Chris-Boe (approved):
Thanks for catching this!
This should only happen when going directly to the page /request-access/id# because the dashboard sends along these state values (in props.location.state) with the button click. We don't necessarily need to pass along any of this state though, we should replace this state passing with an API call on the page so it can handle deep linking.

PR #753, opened 2021-11-21
Create tests and refactor add property manager view

### LINKED ISSUE #744: Create system test for adding a property manager
**Purpose**
Create system test for adding a property manager

### LINKED ISSUE #747: Assigning a manager to a property when creating a manager removes all other assigned managers
**Describe the bug**
Assigning a manager to a property when creating a manager removes all other assigned managers

**To Reproduce**
1. Go to add manager
2. fill in form and assign to a property that has other managers assigned
3. submit form
4. visit show page of the property and notice that only one manager is assigned

I actually can't verify that this is happening, but I'm fairly certain it is.

The reason I can't verify is that when adding a manager with an assigned property is broken..

This code here probably does not do what was intended:

https://github.com/codeforpdx/dwellingly-app/blob/f490c43e6bbdf8f1dedede2a86a6cf73d754225b/src/views/AddPropertyManager/index.jsx#L81-L88

a manager should not be assigned to a property through the property endpoint. The property endpoint will completely replace all the managers with the new array of managers, hence the reason it takes an array instead of a single manager.

To add a manager to a property: This should be done through:
A. either a different endpoint
B. Preferably through the manager create endpoint when creating a manager.

I've added a backend ticket to add a new user/invite method that is specifically for managers so that we can create managers and assign them to zero, one or more properties. 

This ticket is blocked until issue #746 and all other issues needed to get this endpoint up and running.

### LINKED ISSUE #752: Remove hard-coded password
**Describe the bug**
Remove hard-coded password..
A password is not required to create a user.

https://github.com/codeforpdx/dwellingly-app/blob/dfb050dee594e63f2f9378fa8ffb8c14286ed5f5/src/views/AddPropertyManager/index.jsx#L71

NickSchimek:
 - Create tests and refactor add property manager view
     - Fixes #744 Create system tests for adding a property manager.
     - Fixes #747 Assign properties to a new property manager without removing other assignments for the property.
     - Fixes #752 Remove hard-coded password
     

I'm not so sure I like the second commit vvv. 
- Retrieve properties only if users intends to search
   - Retrieves the properties prior to the first search instead of on page load.
   - This decreases the amount of data and requests that are made to load the initial page.

Is a `npm install` necessary? **YES**
Any special requirements to test? NO

### DIFF
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -18,6 +18,24 @@ class InvitationsController < Devise::InvitationsController
   end
 
   def invite_params
-    params.require(:invitation).permit(:firstName, :lastName, :phone, :email, :type, :role)
+    params.require(:invitation).permit(shared_params + resource_params)
+  end
+
+  def shared_params
+    [:firstName, :lastName, :phone, :email, :type, :role]
+  end
+
+  def resource_params
+    case resource_class.new
+    when PropertyManager
+      [{ property_ids: [] }]
+    else
+      []
+    end
+  end
+
+  # Overwrite Devise::Invitations resource_class method
+  def resource_class
+    params.dig(:invitation, :type).constantize
   end
 end
--- a/app/javascript/components/src/views/AddPropertyManager/index.jsx
+++ b/app/javascript/components/src/views/AddPropertyManager/index.jsx
@@ -9,8 +9,8 @@ import Button from "../components/Button";
 import Modal from '../components/Modal';
 import useMountEffect from '../../utils/useMountEffect';
 import './styles/index.scss';
-import UserType from '../../Enums/UserType';
 import RoleEnum from '../../Enums/RoleEnum';
+import UserType from '../../Enums/UserType';
 
 const validationSchema = Yup.object().shape({
   firstName: Yup.string()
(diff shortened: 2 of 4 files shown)


src/views/AddPropertyManager/index.jsx
```diff
@@ -39,57 +37,55 @@ const AddPropertyManager = () => {
...
+    }
   };
 
   const handleFormSubmit = (data) => {
-    let body = {
+    const payload = {
       ...data,
-      password: 'changeAt1stLogin!',
-      confirmPassword: 'changeAt1stLogin!',
-      type: UserType.PROPERTY_MANAGER,
```
Chris-Boe: Don't we need the "type" parameter since we are moving to that from the Role pattern?

src/views/AddPropertyManager/index.jsx
```diff
@@ -39,57 +37,55 @@ const AddPropertyManager = () => {
   const [propertyOptions, setPropertyOptions] = useState([]);
   const [propertySearchResults, setPropertySearchResults] = useState([]);
   const [showAddProperty, setShowAddProperty] = useState(false);
-
-  useMountEffect(() => getProperties());
```
Chris-Boe: The move to getting properties "onBlur - but only once" doesn't seem to work and I'm not sure it's necessary to remove the call from the loading of the page. It's asynchronous so the loading of the page doesn't wait for the property call to complete, but if we put it in the onBlur of the formik component then the user has to wait after they clicked on the search field for the results to come back. We also want the list to appear ASAP when they click even without searching in the field.
I think we can switch it back to querying the properties on page load and take the "propertyRefresh" state out as well.

Chris-Boe (changes requested):

Chris-Boe:
This'll need a rebase too after the other PR's that have been merged

src/views/AddPropertyManager/index.jsx
NickSchimek: > I think we can switch it back to querying the properties on page load and take the "propertyRefresh" state out as well.

ok

src/views/AddPropertyManager/index.jsx
NickSchimek: Not anymore. The endpoint this uses was changed to `context.apiCall('post', /user/invite/property_manager` which only creates a property manager.

PR #766, opened 2021-12-02
Bug/765 archived managers wrong status

### LINKED ISSUE #765: Archived managers do not have correct status in Property Manager List
**Describe the bug**
https://github.com/codeforpdx/dwellingly-app/blob/253a96bb0e2061055cece637e4c69eabc1fd1125/src/views/PropertyManagerList/index.jsx#L25-L32

I believe the logic here should be `&&` instead of `||` on line 25

Additionally the function that is used to format the data will return a string of Invalid Date which is truthy if it cannot format the date.

It would be good to extract this logic out to a function and unit test it in isolation to ensure the correct description is returned for the possible inputs.

**To Reproduce**
1. Create an archived property manager that has a last active time stamp
2. Go to manage/managers

**Expected behavior**
Should see archived status

jumperck:
This PR is solving the problem when showing the status of the Property Manager.
It also solves a build problem due to node version specification.

### DIFF
--- a/src/utils/date/index.js
+++ b/src/utils/date/index.js
@@ -1,0 +1,10 @@
+const formatDate = dateString => {
+  if(isValidDate(dateString))
+    return new Date(dateString + ' UTC').toLocaleString()
+   else 
+    return null
+};
+
+const isValidDate = d => isNaN(Date.parse(d)) === false
+
+export { formatDate, isValidDate };
--- a/.github/workflows/system_test.yml
+++ b/.github/workflows/system_test.yml
@@ -21,6 +21,11 @@ jobs:
   system_test:
 
     runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [14.x]
+
     services:
       postgres:
         image: postgres
@@ -38,6 +43,10 @@ jobs:
       - uses: actions/checkout@v2
         with:
           repository: codeforpdx/dwellinglybackend
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
       - uses: actions/setup-python@v2
         with:
           python-version: '3.9.6'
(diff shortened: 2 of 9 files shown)


src/utils/date/index.js
```diff
@@ -0,0 +1,10 @@
+const formatDate = dateString => {
+  if(isValidDate(dateString))
+    return new Date(dateString + ' UTC').toLocaleString()
```
NickSchimek, line 3: The date is being modified after its check for validity. What if the act of modification causes the date to be invalid?

There should be a contract between the backend and frontend on the date format, but if that contract is inadvertently broken then I think we would like isValidDate to return false if appending UTC to the date causes it to be invalid.

NickSchimek (commented):
Thanks for the tests! I've got one comment below. Let me know what you think.

NickSchimek:
Thanks for all the work you put into this. I think it looks good. Let's get this merged.

NickSchimek (approved):

src/utils/date/index.js
kauereinbold: I agree with you Nick, we might have a problem if the contract is broken for some reason.
One suggestion is to change how we expose the dates through the API. I looked on the Manager list http call and noticed that the date is being transformed on the backend. 
image
image

image

I think we just need to send the date in ISO format and leave the front end handle it.
The only contract we will follow is that the backend sends a UTC date.

What do you think?

NickSchimek: > I think we just need to send the date in ISO format and leave the front end handle it.
> The only contract we will follow is that the backend sends a UTC date.
> 
> What do you think?

Yes, I've been wanting to do that for a long time.

We're going to be combining the frontend and backend into one app/repository. See PR #769

It doesn't do any date transformations like the old backend. And sends all dates in ISO format.

I haven't implemented anything that formats the dates on the frontend, so if someone is wanted to they could implement that.

PR #773, opened 2021-12-20
Prevent archived or unauthorized users from logging in

NickSchimek:
Is a `yarn install` necessary? NO
Any special requirements to test? NO

### DIFF
--- a/app/javascript/components/src/Auth.js
+++ b/app/javascript/components/src/Auth.js
@@ -23,8 +23,8 @@ export const auth = {
         }
       })
       .catch((error) => {
-        Toast("Failure signing in", "error");
-        return Promise.reject(error);
+        Toast(error.response.data, "error");
+        return null
       });
   },
   async signout() {
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -3,9 +3,10 @@
 class SessionsController < Devise::SessionsController
   # before_action :configure_sign_in_params, only: [:create]
 
-  # GET /resource/sign_in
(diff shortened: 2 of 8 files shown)


app/javascript/components/src/Auth.js
```diff
@@ -23,8 +23,8 @@ export const auth = {
         }
       })
       .catch((error) => {
-        Toast("Failure signing in", "error");
-        return Promise.reject(error);
+        Toast(error.response.data, "error");
+        return null
```
aedwardg, line 27: I'm curious about changing this to return null. What's the rationale behind it?

aedwardg (approved):
Looks good at first glance :rocket:

app/javascript/components/src/Auth.js
NickSchimek: ah good question. 

The success path returns an object, and the calling method is expecting an object or nil to be returned.
When a promise is returned it causes an unhandled exception error to occur. 

https://github.com/codeforpdx/dwellingly-app/blob/b27e4fe3bda54bbbae8082c58ddde2fa6fb8d9fa/app/javascript/components/App.jsx#L70-L72

The else condition was removed in favor of using the response message which is more specific.

Alternatively, I guess I could've returned the promise, catch it, and display the toast message, but that seems a bit too much since there is nothing else that needs to be done other than displaying an error message.

NickSchimek: The communication between these two methods seem kinda awkward. We probably don't need two methods here. I'd probably like it if it was one method since it's only used in one place. Got to leave something to improve upon later I guess.

PR #778, opened 2022-01-09
Clean up unused/obsolete code left over from scaffolding

NickSchimek:


yunnie:
I'm still a little unfamiliar with Rails. Is `respond_to` no longer necessary because we don't need to respond with multiple formats? Such as `html` and `json`? Since we only need to respond with `json` we don't need `respond_to`?

yunnie (approved):
I'm still learning more and more about Rails on a regular basis. But the changes make the code more readable. I was initially unclear why we were returning any HTML, and now that the `format.html` is now gone, the code is easier to understand.

NickSchimek:
> I'm still a little unfamiliar with Rails. Is respond_to no longer necessary because we don't need to respond with multiple formats? Such as html and json? Since we only need to respond with json we don't need respond_to?

yeah, exactly. respond_to allows you to specify how to respond to different request formats. It can also be used to explicitly state that we only respond to the specified request formats.

https://api.rubyonrails.org/classes/ActionController/MimeResponds.html#method-i-respond_to

PR #780, opened 2022-01-14
Create property error msg

### LINKED ISSUE #740: Creating property with a name that already exists does not display the error message
**Describe the bug**
A property name is unique.
When creating a property with a name that already exists. The error message is not displayed. Instead the user sees that response status code, instead of the responses error message.

**To Reproduce**
1. Go to add property
2. fill in the form and provide a property name with one that already exists in the system
3. submit form

**Expected behavior**
I'd expect an error message that says the property name already exists.

**Screenshots**
Screen Shot 2021-11-14 at 8 44 43 AM

yunnie:
Closes #740

- [x ] There aren't any unnecessary commits or files changed

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -45,8 +45,8 @@ group :development, :test do
   gem 'rubocop', '~> 1.23', require: false
   gem 'rubocop-rails', '~> 2.12', '>= 2.12.4', require: false
   gem 'rubocop-rspec', '~> 2.6', require: false
-  gem 'sqlite3', '~> 1.4', '>= 1.4.2', require: false
   gem 'shoulda-matchers', '~>5.1.0'
+  gem 'sqlite3', '~> 1.4', '>= 1.4.2', require: false
 end
 
 group :development do
--- a/Gemfile
+++ b/Gemfile
@@ -46,6 +46,7 @@ group :development, :test do
   gem 'rubocop-rails', '~> 2.12', '>= 2.12.4', require: false
   gem 'rubocop-rspec', '~> 2.6', require: false
   gem 'sqlite3', '~> 1.4', '>= 1.4.2', require: false
+  gem 'shoulda-matchers', '~>5.1.0'
 end
 
 group :development do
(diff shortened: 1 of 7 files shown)


yunnie:
I'm looking for feedback. I can either convert the `error.response.data` from `Axios` to a string in the `App.jsx` file. This would change the behavior for error messages for the entire app. That code is currently commented out.

Or I can handle the `error.response.data` from the `AddProperty/index.jsx` file. In this case, two Toast Error messages will be displayed. One from `App.jsx` which will give the HTTP error response code, and the other which will read `"name": ["already taken"]` Personally, I think having some useful feedback like "name already taken" is a bit more helpful to the user.

Chris-Boe:
> I'm looking for feedback. I can either convert the `error.response.data` from `Axios` to a string in the `App.jsx` file. This would change the behavior for error messages for the entire app. That code is currently commented out.
> 
> Or I can handle the `error.response.data` from the `AddProperty/index.jsx` file. In this case, two Toast Error messages will be displayed. One from `App.jsx` which will give the HTTP error response code, and the other which will read `"name": ["already taken"]` Personally, I think having some useful feedback like "name already taken" is a bit more helpful to the user.

Thanks for doing this! I usually prefer the components themselves to handle 400 level errors like this, so they can convert the Toast message to something readable, insert additional data, or have a side-effect to recover from the error. This also would've been the only way to really have a good experience with the old API because different API's returned different responses when there was an error. Is that updated now with Ruby so that all API's return a consistent structure for an error?

If that's the case, I think we can have the global Axios handler prettify and show a toast with the error message while still handing the error message and status code down to the component level so the component can add extra functionality in a catch statement.

yunnie:
I'm a Ruby and Rails noob, but I would think that the error messages from Rails would be pretty consistent in structure and form. I just can't see the Rails core team allowing inconsistently structured error messages. But I would defer to someone with more experience with Rails.

yunnie:
@Chris-Boe Thank you for the suggestion on how to display the back-end validation error in the form via Formik. I ended up adding a parameter to the `apiCall` function in the `App.jsx` file for Formik's `setFieldError`. I kept the Toast message, just so that people know that there is an error. I also made sure that the form does not reset if there is a backend validation error.

NickSchimek (approved):

PR #787, opened 2022-02-07
Move js specs to spec dir and configure ci to run js tests, fix signup

### LINKED ISSUE #786: Sign up form not working
Sign up form is not working

NickSchimek:
- Fix signup form
Fixes #786 
- Move Javascript specs to spec dir
- Configure CI to run JS specs
- Move js dev deps into dev deps section


Chris-Boe (dismissed):

app/controllers/registrations_controller.rb
```diff
@@ -3,7 +3,10 @@
 class RegistrationsController < Devise::RegistrationsController
   respond_to :json
 
-  # before_action :configure_sign_up_params, only: [:create]
+  # rubocop:disable Rails/LexicallyScopedActionFilter
+  before_action :configure_sign_up_params, only: [:create]
+  # rubocop:enable Rails/LexicallyScopedActionFilter
```
aedwardg, line 8: I'm curious why we need to disable rubocop and then enable it again here.

spec/javascript/components/src/views/ForgotPassword.spec.js
```diff
@@ -0,0 +1,42 @@
...
+
+  it("should render without errors", async () => {
+    expect(
+      screen.getByText("REQUEST EMAIL PASSWORD RESET")
+    ).toBeInTheDocument();
+  });
+
+  it("should make an API call when form is submitted", async () => {
+    fireEvent.change(screen.getByPlaceholderText(/Please enter your email address/), mockInputEmailEvent);
+    const button = view.container.querySelector("button");
```
aedwardg, line 36: non-blocking: I think it's an anti-pattern to use `querySelector` in React testing instead of some of the other React Testing Library methods, but this can be changed later since it's from a previously written test.

aedwardg (dismissed):
Looks good to me! 🚀

app/controllers/registrations_controller.rb
NickSchimek: This is because the create action is not defined in this file. It is inherited. And rubocop complains about that because it thinks its not within scope, but it is.

NickSchimek: https://www.rubydoc.info/gems/rubocop/0.58.0/RuboCop/Cop/Rails/LexicallyScopedActionFilter

NickSchimek: I guess rubocop wants this before_action defined on the superclass where the create method exists, but that class is not in our application so we can't do that. So I ignore it.

spec/javascript/components/src/views/ForgotPassword.spec.js
NickSchimek: Thanks, I didn't know about that. Yeah, mainly renaming test files, and making the minimal changes to get them running again.

PR #808, opened 2022-03-06
Setup `bin/dev` file to start server and compile js

NickSchimek:
## Description

Sets up a `bin/dev` file to start the rails server and any other required systems needed for development.

As of now this starts the Rails server and the webpack server so that javascript changes are automatically compiled and reloaded on change.

@Chris-Boe and @kgottfri can you pull this down and approve the PR if it works for you?

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,0 +1,2 @@
+web: bin/rails server -p 3000
+webpack: bin/webpack-dev-server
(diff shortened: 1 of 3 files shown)


kgottfri (dismissed):

Chris-Boe (changes requested):
🤔

image

NickSchimek:
It seems this is a common windows issue when running a bash script.

Using `dos2unix` is mentioned when I google `bin/dev '\r' command not found`

using bin/dev isn't necessary. So maybe I will just update the doc with instructions for booting up the servers with the way we had it and this way as an alternate way.

PR #813, opened 2022-03-20
Make 'add property manager' optional in AddProperty component

### LINKED ISSUE #745: Unable to search and assign property manager when creating a property via the create manager page
**Describe the bug**
Assigning a property manager on the Add Property Modal within the the add Property Manager page does not allow the user to filter the results, and does not allow the user to scroll through the list of results.

**To Reproduce**

Prior data needed:
There must be a lot of property managers in the system, enough to fill your screen height with results.

1. Go to `/add/manager` endpoint
2. Click on `Create New Property`
3. Click in the Search Property Managers search panel
4. Notice it does not allow to search and immediately displays the entire list without the ability to scroll.

**Expected behavior**

**Screenshots**
Screen Shot 2021-11-15 at 3 03 20 PM

 - OS: macos
 - Browser Chrome
 - Version 94

# Follow up Question

Should the assign property manager even be visible within this context?

If your creating a manager, and find yourself needing to create a new property, then I would assume you don't need to add existing managers to this property. If that's the case they can probably create the property first or add after creating the current PM?

I checked the figma docs, but I couldn't find the page where we add a property manager, and it has this modal open. 

Screen Shot 2021-11-15 at 3 13 15 PM

yunnie:

Closes #745

- [ x] There isn't any unnecessary commented-out code
- [ x] There aren't any unnecessary commits or files changed

### DIFF
--- a/app/javascript/components/src/views/AddTenant/index.jsx
+++ b/app/javascript/components/src/views/AddTenant/index.jsx
@@ -447,6 +447,7 @@ const AddTenant = () => {
                 showPageTitle={false}
                 handleCancel={closePropertyModal}
                 afterCreate={closeAndSetProperty}
+                showAssignPropManagers={true}
               />
             }
             hasButtons={false}
--- a/app/javascript/components/src/views/AddTenant/index.jsx
+++ b/app/javascript/components/src/views/AddTenant/index.jsx
@@ -448,6 +448,7 @@ const AddTenant = () => {
                 handleCancel={closePropertyModal}
                 afterCreate={closeAndSetProperty}
                 showAssignPropManagers={true}
+                propManagerSearchPanelWidth={200}
               />
             }
             hasButtons={false}
--- a/app/javascript/components/src/views/AddTenant/index.jsx
+++ b/app/javascript/components/src/views/AddTenant/index.jsx
@@ -447,8 +447,7 @@ const AddTenant = () => {
                 showPageTitle={false}
                 handleCancel={closePropertyModal}
                 afterCreate={closeAndSetProperty}
-                showAssignPropManagers={true}
-                propManagerSearchPanelWidth={200}
+                showAssignPropManagers={false}
               />
             }
             hasButtons={false}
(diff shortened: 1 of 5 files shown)


app/javascript/components/src/views/AddTenant/index.jsx
```diff
@@ -447,6 +447,7 @@ const AddTenant = () => {
                 showPageTitle={false}
                 handleCancel={closePropertyModal}
                 afterCreate={closeAndSetProperty}
+                showAssignPropManagers={true}
```
Chris-Boe: Setting this to true for the modal to show the property manager selection still breaks the functionality when the list is too long. Should we just switch it to false if we don't need it here as well?

Chris-Boe (changes requested):

app/javascript/components/src/views/AddTenant/index.jsx
yunnie: The width of the Property Manager search field has a fixed with in `AddProperty` component. Is there a way to adjust the size of the `SearchPanel` so that it takes in the appropriate size automatically? Or we could make the width an input.

yunnie: I added a size option to the `SearchPanel` in the `AddProperty` component. So the property manager search now fits in the modal.

Chris-Boe: The width control helps but it's the vertical size of the dropdown being off the page that's the issue:
image

Chris-Boe: I think we can limit the size across the board and make it scrollable to fix it

Chris-Boe (changes requested):

app/javascript/components/src/views/AddTenant/index.jsx
yunnie: I ended up using your original suggestion and removed the "Assign Property Manager". I tried setting a max height, but no matter what height I used, it always ended up pushing the "save" and "cancel" buttons off the screen.

NickSchimek (approved):

Issue #807, opened 2022-02-26
Updating Property Managers of a Property does not work


Chris-Boe:
**Describe the bug**
Attempting to edit a Property and add a property manager does not apply the change correctly. There are two issues here:
* UI submits the request to update the property and takes the UI out of Edit mode as soon as you click a property manager from the dropdown when it should only update the property when the save button is clicked
* Backend does not use the propertyManagerIDs request parameter to update the Property's assigned Property Managers

**To Reproduce**
1. Go to Properties list
2. Click on a property to edit it
3. Click edit using the pencil icon
4. Select a property manager from the search bar dropdown (population of the dropdown fixed by #801 )
5. See that the property manager was not correctly assigned

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #818, opened 2022-07-04
Issues 714, 807 Fixes

### LINKED ISSUE #714: Allow editing of Property, Unit, Lease, and JOIN Staff for existing tenants
**Purpose**
What will this feature provide?
Allows admins to edit the property, lease, and unit for existing tenants from the Manage -> Tenant screen

**Proposed Solution**

**Acceptance Criteria**
- [x] Property should be able to be picked from a populated dropdown when in Edit mode
- [x] Unit should be editable
- [ ] Lease should be editable
- [x] When Save is clicked, these fields are also sent to the backend and the frontend reflects the update
- [x] JOIN Staff should only be editable when in Edit mode and should be saved when Save is clicked (should we move Save button down so it looks like it encapsulates both sections?)

Paste a screenshot of the feature from the FIGMA mockups or cite the page number.

**Additional context**

Chris-Boe:
Fixes #714 and #807 

Is a `yarn install` necessary? n


Chris-Boe:
@NickSchimek The changes I made to the tenant controller somehow broke the assignment of staff to tenants in the PATCH requests from the dashboard. The commit above "Fix Dashboard tenant staff assignment" changes it to a PUT just to get it to work, but I think we can toss the commit if I get your help to figure out wtf is going on with the Ruby stuff 😅

NickSchimek:
@Chris-Boe `PUT` and `PATCH` requests are treated the same in Rails so I don't think that's it. I see other changes in that commit and a lot of changes elsewhere. So I don't think it makes a difference wether put or patch is used.

Chris-Boe:
> @Chris-Boe `PUT` and `PATCH` requests are treated the same in Rails so I don't think that's it. I see other changes in that commit and a lot of changes elsewhere. So I don't think it makes a difference wether put or patch is used.

Reversing just the changes in the tenants controller fixes the dashboard PATCH request and then broke the PUT request that was originally broken so it seems it's something about that update controller method both requests go through? We can investigate wednesday maybe or just leave it with the PUT fix. It's more data we're sending but I don't think it's a big deal

cypress/integration/tenants/add_tenant.spec.js
```diff
@@ -22,13 +22,17 @@ describe('Add Tenant', function () {
   })
 
   it('creates a tenant with a lease', function () {
+    cy.visit('add/property')
+    cy.wait(500)
+    cy.fillInPropertyForm()
+    cy.get('button').contains('SAVE').click()
+    cy.wait(500)
```
NickSchimek: We shouldn't use the UI to create a property if its needed for this test.

Rather than doing this. 

We can update line 7 to create a property.

```js
    cy.appFactories([
      ['create', 'staff', {email: 'staff@dwellingly.org', firstName: 'Franky', lastName: 'Bob'} ],
      ['create', 'property, {name: 'Property Name', address: 'address', ... } ]
    ])
```

NickSchimek: You don't need to pass any attributes. Only ones you want to override. Otherwise this file will add all the attributes automatically when a property is created.

`spec/factories/properties.rb`

cypress/integration/tenants/add_tenant.spec.js
```diff
@@ -22,13 +22,17 @@ describe('Add Tenant', function () {
...
     cy.visit('/add/tenant')
     cy.fillInTenantForm()
-    cy.contains('Create New Property').click()
-    cy.fillInPropertyForm()
-    cy.get('form.add-property__form-container').contains('SAVE').click()
-    cy.get('form.add-property__form-container').should('not.exist')
-    cy.contains('The Tenants').should('exist')
+
+    cy.get('input[title="Search Properties"]').first().type('The')
+    cy.get('.styles-module_resultItemLabel__rcaln:contains(The Tenants)').click()
```
NickSchimek, line 30: It looks like this was testing you could create a property when adding a tenant.
This is now testing you can find and select a property when adding a tenant.

I think this is better 👍

NickSchimek (approved):
Very nice refactoring!

PR #831, opened 2022-07-19
Fixes #799 Accept user invitation

### LINKED ISSUE #799: Unable to accept invitation in email
**Describe the bug**

**To Reproduce**
1. Go to Add PropertyManager
2. Fill in details and submit
3. Grab the invitation token either from the db or from the server log

You might be able to use this to see the error

```
http://localhost:3000/api/users/invitation/accept?invitation_token=BJvxUbgDqgwesobk5yWi
```

Screen Shot 2022-02-21 at 1 23 27 PM

I'd expect the invitation to be accepted and automatically logged in.

NickSchimek:
We render a Set Password form when the user accepts the invitation from their mailbox.
After entering a password they are redirected to the login page.

We could optionally log them in if we choose and display the dashboard.

Fixes #799

Is a `yarn install` necessary? no

### DIFF
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -1,6 +1,5 @@
 class InvitationsController < Devise::InvitationsController
-  before_action :authorize_user
-  after_action :verify_authorized
+  before_action :authorize_user, except: [:edit, :update] # rubocop:disable Rails/LexicallyScopedActionFilter
 
   def create
     self.resource = invite_resource
@@ -41,8 +40,28 @@ class InvitationsController < Devise::InvitationsController
     end
   end
 
-  # Overwrite Devise::Invitations resource_class method
+  # Overwrite Devise::Invitations resource_class method so that we can create the correct UserType.
+  # Alternatively we could update our routes so that we have a different route for each type of
+  # user. This will also require frontend updates to use the new routes, rather than sending
+  # the type parameter in the payload.
   def resource_class
-    params.dig(:invitation, :type).constantize
+    return params.dig(:invitation, :type).constantize if params.dig(:invitation, :type).present?
+
+    super
   end
+
+  # This method is called to validate the token. It redirects on an invalid or blank token.
+  # rubocop:disable Layout/MultilineOperationIndentation
+  # rubocop:disable Lint/AssignmentInCondition
+  # rubocop:disable Rails/DynamicFindBy
+  def resource_from_invitation_token
+    unless params[:invitation_token] &&
+      self.resource = resource_class.find_by_invitation_token(params[:invitation_token], true)
+
+      redirect_to root_path(error: :invalid_invitation_token)
+    end
+  end
+  # rubocop:enable Layout/MultilineOperationIndentation
+  # rubocop:enable Lint/AssignmentInCondition
+  # rubocop:enable Rails/DynamicFindBy
 end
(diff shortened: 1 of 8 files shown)


app/controllers/invitations_controller.rb
```diff
@@ -36,6 +36,17 @@ def resource_params
 
   # Overwrite Devise::Invitations resource_class method
   def resource_class
-    params.dig(:invitation, :type).constantize
+    return params.dig(:invitation, :type).constantize if params.dig(:invitation, :type).present?
+
+    User
```
aedwardg: Could this be `super` instead of being explicitly `User`?
I think in the end `super` might have the exact same result, but leave open flexibility for future use.  Thoughts?

app/controllers/invitations_controller.rb
```diff
@@ -36,6 +36,17 @@ def resource_params
...
   end
+
+  # This method is called to validate the token. It can be overwritten.
+  # def resource_from_invitation_token
+  #   unless params[:invitation_token] &&
+  #     self.resource = resource_class.find_by_invitation_token(params[:invitation_token], true)
+  #     set_flash_message(:alert, :invitation_token_invalid) if is_flashing_format?
+  #     redirect_to after_sign_out_path_for(resource_name)
+  #   end
+  # end
```
aedwardg: Do we need this comment?

aedwardg (dismissed):
Nice, this works now!

The password setting page could use some styling if we decide to use it 😂 but otherwise, this is good to go.

One question I do have is about the log-in flow.  Currently, we have a system where people who try to create an account from the log-in page need to be approved on the admin dashboard. Do we want the same thing here, or are we assuming that if they were invited this way that they are already "pre-approved" to log in?

app/controllers/invitations_controller.rb
NickSchimek: > Could this be super instead of being explicitly User?

Yes, that's what I wanted to do, but super will return an UnauthorizedUser class instead of User because of the route I created. Let me see if I can fix it.

With a little more work we could remove the need for this method altogether. But that will also require frontend updates as well, which at the time this was created I wasn't willing to do, but I am now.

Ideally we can get some tests written to cover the current correct behavior, and then make the updates to remove this method. I will get an issue created after this is merged, and maybe someone will grab it.

app/controllers/invitations_controller.rb
NickSchimek: Yes, I want to leave this here, because we do need to override this method. I'm going to update the comment. Once this is merged in I'm going to get a new issue created so that we can get this fixed.

NickSchimek:
> The password setting page could use some styling if we decide to use it 

Yeah, we need to add some styling if we're using this Rails view. If someone wants to make the React component for it then that will require some extra work beyond some simple styling. Either way we go I'm fine with. I rather use the simple rails view, rather than booting up the react app to set a password.

> Do we want the same thing here, or are we assuming that if they were invited this way that they are already "pre-approved"

Yeah, if a user is invited (which can only be done by an admin) then they are an authorized user at the time of being invited. So no approval is needed.

app/controllers/invitations_controller.rb
NickSchimek: Okay, I ended up hacking something together so that an error message is shown on an invalid token.

PR #832, opened 2022-07-23
Update `lastActive` when user signs in

### LINKED ISSUE #829: Property managers seem to show pending even when they're active
**Describe the bug**
Property managers seem to show pending even when they're active

**To Reproduce**
1. Go to Property managers

**Expected behavior**
Status should be accurate

**Screenshots**
Screen Shot 2022-07-18 at 18 04 58

aedwardg:
closes #829 

The front end was rendering the "Last Usage" column based on the user's `lastActive` status, but this field in the database was never getting updated. Since we're not using devise trackable (which would automatically update `last_sign_in_at`, `current_sign_in_at`, etc.) we need to update this value ourselves.

Note, this saves the datetime in the db in UTC time.  I don't think we alter that for the front-end, so it may look off by a day if checked at certain times.

Screenshot
image

Is a `yarn install` necessary? ❌ 
- Sign up a new user
- Sign in as admin and approve them as Property Manager
- See that they are still marked as Pending
- Sign in as the new user
- Sign back in as admin to see "Last Usage" column updated.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -16,6 +16,7 @@ class SessionsController < Devise::SessionsController
 
     if resource.valid_password?(params[:password])
       sign_in :user, resource
+      resource.update!(lastActive: Time.current)
       return render json: resource.to_json
     end
 
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -16,7 +16,7 @@ class SessionsController < Devise::SessionsController
 
     if resource.valid_password?(params[:password])
       sign_in :user, resource
-      resource.update!(lastActive: Time.current)
+      resource.update_column(:lastActive, Time.current) # rubocop:disable Rails/SkipsModelValidations
       return render json: resource.to_json
     end
 
--- a/spec/requests/sessions_spec.rb
+++ b/spec/requests/sessions_spec.rb
@@ -1,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.describe "/api/resource", type: :request do
+  describe "POST /sign_in" do
+    it "updates the user's lastActive timestamp on sign in" do
+      property_manager = create(:property_manager)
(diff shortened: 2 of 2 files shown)


aedwardg:
@NickSchimek Open to any thoughts on if/how we want to test this.

app/controllers/sessions_controller.rb
```diff
@@ -16,6 +16,7 @@ def create
 
     if resource.valid_password?(params[:password])
       sign_in :user, resource
+      resource.update!(lastActive: Time.current)
```
NickSchimek: I would probably use `update_column` here instead of `update!`, unless there is a reason we would want to run callbacks and validations to update this field?

NickSchimek:
> @NickSchimek Open to any thoughts on if/how we want to test this.

A rails request spec.
When a property_manager signs in. 
We update lastActive and respond successfully with a serialized property manager. 
Check serialization (lastActive) is correct.

app/controllers/sessions_controller.rb
NickSchimek: It looks like this is only used for PropertyManagers, I guess there is no harm in updating it for all types of users.

I believe this is more of a last_sign_in rather than a last_active timestamp, which might be a little bit different than how it was originally intended, but I think that is ok. If it needs to change later we can do that.

aedwardg:
@NickSchimek I just tried using `update_column`, but rubocop doesn't like it. Should I change it back to `update!` ?

NickSchimek:
Oh we can turn that cop off, or disable it for that one line. Your choice.

The spec looks good. Just need to remove the puts

aedwardg:
>Oh we can turn that cop off, or disable it for that one line. Your choice.

I decided to only disable it for that line.  I think in the long run, it's better to have it, especially if people less familiar with Rails contribute and see `update_column` here.

NickSchimek (approved):
Thanks!

PR #833, opened 2022-07-23
use math.div for sass division

aedwardg:
No existing issue to close.

Updates `emify` sass function to use `math.div` instead of `/` for division.
This gets rid of the following deprecation warning when starting up the app:

image

If we upgraded the version of sass we use, it could be changed to use the [calculations syntax](https://sass-lang.com/documentation/values/calculations) `calc($target / $context)` and we wouldn't need to import the math module, but I'm not sure if we can safely upgrade sass or if `sass-rails` depends on the current version.

Is a `yarn install` necessary? ❌ 
Any special requirements to test? ❌ 

### DIFF
--- a/app/javascript/components/src/views/components/TicketModal/components/Card/styles/index.scss
+++ b/app/javascript/components/src/views/components/TicketModal/components/Card/styles/index.scss
@@ -1,3 +1,4 @@
+@use "sass:math";
 @import "../../../../../../scss/variables.scss";
 
 
@@ -15,7 +16,7 @@ $width__wrapper: 800px;
 
 /* F U N C T I O N S */
 @function emify($target, $context) {
-  @return ($target/$context) * 1em;
+  @return math.div($target, $context) * 1em;
 }
 /* end functions */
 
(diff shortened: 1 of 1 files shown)


NickSchimek:
> If we upgraded the version of sass we use, it could be changed to use the [calculations syntax](https://sass-lang.com/documentation/values/calculations) calc($target / $context) and we wouldn't need to import the math module, but I'm not sure if we can safely upgrade sass or if sass-rails depends on the current version.

The library that installs the sass package is the `rails/webpacker` npm package. This package is now retired because there are better options out there now. It is on my todo list to replace this when we upgrade.

https://github.com/rails/webpacker/blob/4c5d13a2aedd3e63e1cdadf46575bc832a126235/package.json#L45

You can also see this dependency in our package.json file.

Since we do directly depend on the sass package. Then we should add it to our package file. 

Can you add this as well in this PR? run: `yarn add sass@^1.38.0`

With that added our code will not break when we do remove `rails/webpacker` 

I think Longterm goal would be to move these files to the stylesheets directory

NickSchimek (approved):
I think this change is okay. Although I will punt to Chris here.

I'd recommend adding sass to our dependency file (since we directly depend on it) rather than having it indirectly depended on.

The latest changes for `rails/webpacker` no longer includes sass as a dependency...

Chris-Boe (approved):

PR #836, opened 2022-08-02
Provide role info to frontend

NickSchimek:


Chris-Boe (approved):
Just had one question: What is Staff Level?

NickSchimek:
> Just had one question: What is Staff Level?

Returns true if the user is either an admin or staff; otherwise false.

https://github.com/codeforpdx/dwellingly-app/blob/a5cdd95127f3165ed083fdfe392aa3ae5e891082/app/models/user.rb#L18

PR #844, opened 2022-08-13
Fix scoping on notes

### LINKED ISSUE #841: Edit and Delete Note APIs are missing
**Describe the bug**
Either the frontend is hitting the wrong API or the API is missing in the backend. The frontend is currently sending Note Edits to:
PATCH /api/tickets/{ticketId}/notes/{noteId}
and it's coming back with a  500

It's also sending Delete requests to:
DELETE /api/tickets/136/notes/2
and it's coming back with a 500

**To Reproduce**
1. Go to a ticket
2. Add a note
3. Attempt to edit note and save changes
5. Attempt to delete note and save changes
6. See error

**Expected behavior**

**Screenshots**
Edit:
image

Delete:
image

**Additional context**

NickSchimek:
Pass in ticket.notes to the notes policy scope.

Fix scope when its a property manager. A Property manager can only update/destroy their own note.

Fixes #841

Is a `yarn install` necessary?


app/policies/note_policy.rb
```diff
@@ -9,7 +9,7 @@ def resolve
       if user.staff_level?
         scope.all
       else
-        user.properties.property_managers
+        scope.where(user: user)
```
NickSchimek, line 12: I don't remember if a property manager is allowed to edit/destroy any note as long as it was created by one of their co-managers.

If that is the case then this should be `scope.where(user: user.properties.property_managers)`

Chris-Boe (approved):

app/policies/note_policy.rb
Chris-Boe: I think this is fine for now, we can come back to it if we need to

